PR #968, opened 2015-01-29
Use sha2 digest instead of md5 for tar package.

hsbt:
I got offering to use secure digest without MD5. see https://bugs.ruby-lang.org/issues/10793

I replace md5 to sha2 for download page.


stomar:
s/sha2/sha256/

sha2 is unclear since there are differents variants (sha512 is also sha2).

PR #969, opened 2015-02-02
fix typo

gitname:
changed "classes" (plural) to "class's" (singular possessive).


stomar:
Thanks! Merged manually.

(Note there is a difference between the commit message and the description of the PR. The commit message is not the place for verbose explanations.)

PR #995, opened 2015-02-28
Minor fixes (proposal)

kant:


chikamichi:
Hi,

Try to think about providing the affected language in the commit title, like "(es)" for instance. Title which is not really meaningful btw ;)

PR #996, opened 2015-02-28
Minor fix (proposal)

kant:


dabit:
Please don't forget to include the language on the title of Pull Requests and Commit descriptions for future proposals.

An example of a good title for this particular PR would be:

*Minor fixes for index page (es)*

Thanks for the help!

PR #1012, opened 2015-03-03
Update rubies in _config.yml

makimoto:
I update `_config.yml` for new version rubies.

* 2.2.0 => 2.2.1
* 2.0.0-p598 => 2.0.0-p643

@ruby/www-ruby-lang-org please check it.


hsbt:
:+1:

Issue #1019, opened 2015-03-07
Error on Ruby in Twenty Minutes Portugues version


LuizFlavio:
On tutorial Ruby in Twenty Minutes Portugues version there's a error in text.

In the item "Evoluindo para um Anfitrião", the code in text box is:

```
irb(main):024:0> class Anfitriao
irb(main):025:1>   def initialize(nome = "Mundo")
irb(main):026:2>     @nome = nome
irb(main):027:2>   end
irb(main):028:1>   def diz_ola
irb(main):029:2>     puts "Olá #{@nome}!"
irb(main):030:2>   end
irb(main):031:1>   def diz_adeus
irb(main):032:2>     puts "Adeus #{@nome}, volte sempre."
irb(main):033:2>   end
irb(main):034:1> end
=> nil
```

It defines the methods of Anfitriao class as `diz_ola` and `diz_adeus`. But ahead in text, when the object is created with the name João, the methods of class Anfitriao are written as "`h.dizer_ola`" and "`h.dizer_adeus`" instead of "`h.diz_ola`" and "`h.diz_adeus`", generating NoMethodError: undefined method.

Text with the error:

```
irb(main):035:0> h = Anfitriao.new("João")
=> #<Anfitriao:0x16cac @nome="João">
irb(main):036:0> h.dizer_ola
Ola João
=> nil
irb(main):037:0> h.dizer_adeus
Adeus João, volta em breve.
=> nil
```

stomar:
Fixed with #1020.

PR #1020, opened 2015-03-10
Use dizer_ola instead of diz_ola for Anfitriao class (pt)

spk:
see #1019

### DIFF
--- a/pt/documentation/quickstart/2/index.md
+++ b/pt/documentation/quickstart/2/index.md
@@ -111,7 +111,7 @@ irb(main):027:2>   end
 irb(main):028:1>   def diz_ola
 irb(main):029:2>     puts "Olá #{@nome}!"
 irb(main):030:2>   end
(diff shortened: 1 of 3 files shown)


stomar:
Methods are about sending messages, therefore the imperative should be used, IMO. (Which would be `diz`, if I am not mistaken.)

And please note the affected languate in commit messages, like in e.g. `Change method name in example (pt)`.

chikamichi:
@spk didn't know you speak Portugese.

spk:
@chikamichi I don't I thought it was that.
@stomar updated.

chikamichi:
Pas compris. Tu parles pas portugais mais tu fais des commits en portugais ? :monkey_face:

spk:
@chikamichi Tu as compris :exclamation: :see_no_evil:

JuanitoFatas:
@stomar merge this?

PR #1021, opened 2015-03-10
Fix markup errors with carriage return

spk:
see https://validator.nu/?doc=https%3A%2F%2Fwww.ruby-lang.org%2Ffr%2Fnews%2F2014%2F12%2F18%2Fruby-2-2-0-rc1-released%2F


stomar:
Applied with a081d5ae016361adb746a3b1704cdac2012a0fe2.

PR #1036, opened 2015-03-18
Remove 1.9 series from downloads

makimoto:
Supports for ruby 1.9 series are ended now.
So I remove 1.9 series from downloads page.

@ruby/www-ruby-lang-org please review it.


hsbt:
:+1: Nice works.

JuanitoFatas:
:+1:

JuanitoFatas:
Do we need a legacy page for people who cannot upgrade but still need to download 1.9.3?

postmodern:
I think it's dangerous to recommend installing a Ruby that is no longer supported or maintained.

Issue #1074, opened 2015-05-26
www.r-l.org zh_cn status


JuanitoFatas:
Hi I saw zh_cn is [now marked as unmaintained](https://github.com/ruby/www.ruby-lang.org/commit/6c4d74c3fbf0ca69cdce7f438546d38f1ba6b027), so I gathered some people and will guide them to update outdated of `zh_cn` site.

--

@pmq20 @hlcfan @liuzxc @tzwm @happybai @diguage @alex-marmot @itsliupeng
@cassiuschen @1ternal @killernova @VincentZhao @reyesyang @jun1st @springwq
@Huibean @junzh0u @zigzagpig @42thcoder @GaaraLi @shouxian @johnnyhappy365
@aidewoode @Yunfei1982 @Airmacho @cursorzz @isjia @debbbbie @AndyHelix
@Zane5 @VirgilMing @boyishwei @dandananddada @jundechen @mohongshu

Hello everyone! :wave: :wave: :wave: 

First thank you for considering helping out! :heart: Below you can find list of things that need your help. You can [read the wiki](https://github.com/ruby/www.ruby-lang.org/wiki) to get started and take a look some of these examples to see how this works:

- Translate Page example: [Example 1](https://github.com/ruby/www.ruby-lang.org/pull/997), [Example 2](https://github.com/ruby/www.ruby-lang.org/pull/973)
- Update outdated page example: [Example 1](https://github.com/ruby/www.ruby-lang.org/pull/991), [Example 2](https://github.com/ruby/www.ruby-lang.org/pull/913)
- Fix typo example: [Example 1](https://github.com/ruby/www.ruby-lang.org/pull/1059), [Example 2](https://github.com/ruby/www.ruby-lang.org/pull/1016)
- Translate news example: [Example 1](https://github.com/ruby/www.ruby-lang.org/pull/1056), [Example 2](https://github.com/ruby/www.ruby-lang.org/pull/1045)

**Please do leave a comment if you want to work on some of the things to avoid people double work on the same thing.** :sweat_smile:  

### Style of Translation

Please refer to https://github.com/sparanoid/chinese-copywriting-guidelines 

--

#### Pages

~~- Licenses https://www.ruby-lang.org/en/about/license.txt @shouxian~~
- [x] Downloads https://www.ruby-lang.org/en/downloads @hlcfan 
- [x] Documentation https://www.ruby-lang.org/en/documentation @1ternal
[...]

PR #1108, opened 2015-06-03
To ruby from python zh cn

Huibean:
Related #1074


VincentZhao:
I think it should be a markdown file, instead of a html file.
Please see [the en version](https://github.com/ruby/www.ruby-lang.org/blob/master/en/documentation/ruby-from-other-languages/to-ruby-from-python/index.md).

Secondly, no need to change Gemfile.lock

Huibean:
have put the translation in index.md

Gemfile.lock
```diff
@@ -109,3 +109,6 @@ DEPENDENCIES
   spidr (~> 0.4)
   unicorn
   validate-website (~> 0.9)
+
+BUNDLED WITH
+   1.10.1
```
JuanitoFatas: Could you revert this change? Because not all people are using 1.10.x. Thanks.

JuanitoFatas:
As @VincentZhao said, please commit markdown file instead of generated html. Thank you.

stomar:
Please provide a new PR. You must not change Gemfile.lock or _site, only submit the markdown file.

And please use a commit message that conforms to our conventions, like `Translate to-ruby-from-python (zh_cn)`.

PR #1113, opened 2015-06-05
Add RubyBib to the sidebar

chrisseaton:
@hsbt reference 61eaf874269102dd7dedb638a55b7f722afb466f - I didn't realise I needed to change the sidebar template as well - I thought just changing the configuration was enough. This fixes that and should put it into the sidebar.


hsbt:
@chrisseaton :+1: Thanks for your followup.

Issue #1104, opened 2015-06-01
link to shut down website in success stories


jannisch:
There is a link on the success stories page (https://www.ruby-lang.org/de/documentation/success-stories) to www.43things.com which now refers to a domain parking page. According to Wikipedia, "43things.com became read-only on August 15, 2014, and shut down permanently January 1, 2015".

JuanitoFatas:
Could you open a Pull Request to remove it? Thank you.

jannisch:
Yes, I hope it's like you expected!

zzak:
Thanks for your report!

PR #1114, opened 2015-06-05
Remove link to shut down website (all languages)

jannisch:
this patch for issue [#1104](https://github.com/ruby/www.ruby-lang.org/issues/1104) removes the mentioned link to 43things.com and corrects the reference link number of the following ones.


JuanitoFatas:
Looks good to me.

@hsbt?

zzak:
Thank you! But also feeling nostalgic after finding this photo from 43things wikipedia of @drbrain:

43things

PR #1126, opened 2015-06-18
Update index.md

paul59:
In my opinion (as a new Rubyist) it's confusing having :names, @names, names and |name| all in the same small example. This change removes a little confusion.


stomar:
It's common to use the same names (and not `a_name` and `@name`), so you know in which attribute you have to look for (that way you do not have to search for the assignment line). Furthermore, `somename` and `names` are conceptually different.

And please use descriptive commit messages for future PR (see our Wiki page for some hints).

PR #1132, opened 2015-06-27
Improve SSL configuration

stomar:
@hsbt Any objections?

Activating rack-ssl only in production makes serving locally using `rackup` easier. (I couldn't manage to do that with SSL on.) Compared to serving the site using e.g. `jekyll --serve`, local testing could then be done in a more production-like environment, including rewrite rules, rack/jekyll, ...


hsbt:
@stomar :+1: It's good. I think it make simply configuration.

PR #1134, opened 2015-07-03
Add link directly to ruby-core list

schneems:
It is difficult to find the "Welcome to the "ruby-core" mailing list" email to manage your ruby-core list subscription. We should add a link directly to the page for easy access. I made the link body of to this site "subscribe" to keep it short, but we can pick something else. I think "subscribe" is not very accessible to people using screen readers, instead maybe "manage ruby-core list subscription".

### DIFF
--- a/en/community/mailing-lists/index.md
+++ b/en/community/mailing-lists/index.md
@@ -17,7 +17,7 @@ Ruby-Talk
 
 Ruby-Core
 : This list deals with core and implementation topics about Ruby, often
-  used to run patches for review. ([Archives][4])
+  used to run patches for review. ([Subscribe][6] and [Archives][4])
 
 Ruby-Doc
 : This list is for discussing documentation standards and tools for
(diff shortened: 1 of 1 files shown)


schneems:
btw does anyone know why digest mode is disabled for ruby-core?

stomar:
Thanks for the suggestion. I agree that there should be a link to the mailman interface.

But IMO the proposed change makes the page (even?) more confusing. There would be 3 methods of subscribing etc., linked to from 3 different places: the suggested new link to the mailman interface, then the subscription form further down on the page, and then the link to the manual mailing instructions page (which btw _also_ links to lists.ruby-lang.org).

I'll try to come up with a clearer way to structure the page.

matkoniecz:
@stomar Given that this PR was rejected maybe it would be a good idea to close it (and open issue if that is still a problem)?

Open PR from 2015 is not encouraging to submit PRs to this repo.

jcserracampos:
@stomar where we will sing happy birthday to this issue?

PR #1135, opened 2015-07-03
Update links to cache.ruby-lang.org to use HTTPS

konklone:
This updates every link in the site to `cache.ruby-lang.org` to use the HTTPS version, now that [HTTPS is enabled for Ruby downloads](https://bugs.ruby-lang.org/issues/10672#change-53270) (and the configuration is [pretty good too](https://www.ssllabs.com/ssltest/analyze.html?d=cache.ruby-lang.org)).

Thanks very much to @hsbt for working with Fastly to get this turned on for the Ruby project, and to @krues8dr for teaching me the `sed` command to do this easily. :smiley_cat:


konklone:
Thank you again, @hsbt!

konklone:
Is there any deploy schedule, or other hint about when this change might go live?

reedloden:
Does it also make sense to update all the old ftp://ftp.ruby-lang.org links to use the new https://cache.ruby-lang.org URL as well?

konklone:
@reedloden Do the URLs map exactly? Would a find/replace for the protocol/domain do it? Offhand, it looks like it would.

reedloden:
Yeah, looks like they do, and yeah, find/replace would work. http://www.jjj.de/shell/replacepage.html is useful for that. :)

PR #1144, opened 2015-07-15
Markup not found issues

spk:
- Found one 404 on https://www.ruby-lang.org/id/news/2012/10/12/cve-2012-4464-cve-2012-4466/index.html ( masalah keamanan yang mirip ), use English version.
- Other commit is about link generated `:https://bugs.ruby-lang.org/projects/ruby-19/issues?query_id=11`, use same bottom link as other news. (https://www.ruby-lang.org/id/news/2009/01/30/ruby-1-9-1-rilis/)


stomar:
Good catch! Merged with e230018842b841f916112f0d26522936cb9cfbfc. Sorry for the delay.

PR #1146, opened 2015-07-17
Move FTP URLs to HTTPS URLs

konklone:
Inspired by @reedloden's suggestion in https://github.com/ruby/www.ruby-lang.org/pull/1135#issuecomment-122136785, I've:

1. rewritten all the `ftp://ftp.ruby-lang.org` URLs,
2. rewritten all the `http://ftp.ruby-lang.org` URLs...

...to `https://cache.ruby-lang.org` URLs.

This may be more than the project was looking for, but `1.` seems obvious (FTP doesn't offer any security) and `2.` seemed helpful for consistency. I'm happy to file a smaller PR with just `1.` if desired.


konklone:
For my own future reference, these are the commands I used:

```
sed -i 's/ftp:\/\/ftp.ruby-lang.org/https:\/\/cache.ruby-lang.org/g' $(find ./ -type f -name '*.md')
```

```
sed -i 's/http:\/\/ftp.ruby-lang.org/https:\/\/cache.ruby-lang.org/g' $(find ./ -type f -name '*.md')
```

```
sed -i 's/https:\/\/ftp.ruby-lang.org/https:\/\/cache.ruby-lang.org/g' $(find ./ -type f -name '*.md')
```

_config.yml
```diff
@@ -56,14 +56,14 @@ downloads:
       zip: 2ad4eaabfd92d627baffc6c971e4b8987b38c06baf42dc2fc2e05131095499e7
   stable_snapshot:
     url:
-      bz2: https://ftp.ruby-lang.org/pub/ruby/stable-snapshot.tar.bz2
```
reedloden: This may be on purpose since (as I understand it) ftp.ruby-lang.org is the origin server used for cache.ruby-lang.org, and since these and the ones below are snapshots, not sure the CDN makes sense, especially since they change so often.

konklone: Yeah, OK, I added a commit reverting the changes to this file. This removes `3.` from the ticket description, so I'll update that too.

hsbt:
@konklone Thank you for your suggestion. I applied 4106463dc219f1a5df5cd107c987613a9e290f02 and  4fbf9dceb4a8f125ab7122965a61b2a3d2125235 manually.

PR #1166, opened 2015-08-22
Fix broken twitter link in about/website

gmitrev:
The old link is returning a 404 because Jason seems to have changed his twitter handle.


JuanitoFatas:
:+1:

PR #1173, opened 2015-09-02
Fix the year of ConFoo news

yous:
The year of the news should be 2015, right?


JuanitoFatas:
Do we need the redirect rule? /cc @hsbt

hsbt:
@JuanitoFatas It's comfortable.

@yous Thanks!

JuanitoFatas:
@hsbt Thanks! :bow:

PR #1183, opened 2015-09-23
Translate all March 2015 news (id)

meisyal:
As titled.
Please, review my translation @gozali!
Thank you so much.

cc: @stomar


gozali:
+1

PR #1188, opened 2015-10-15
Update 2016 Fukuoka Ruby Award Competition (ko)

yous:
https://github.com/ruby/www.ruby-lang.org/pull/1187/files#r42129269

Removed the Engine Yard prize which is not available now.

See https://github.com/digitalfukuoka/www.ruby-lang.org/commit/de00266466ca602b4cbaa3c4d5bc140689f3de51.


stomar:
Merged manually with 8129f7a5064789dee056afa98cddbb147dd8aff3.

PR #1199, opened 2015-11-04
Add and update translations (pt)

stomar:
This PR supersedes #1131 (rebased, conflicts resolved and small fixes).

PR #1200, opened 2015-11-05
A few more `pt` translation fixes

alanwillms:
Fix two issues as pointed out by @stomar in #1199 

* on the `downloads` page, one complete paragraph is missing (Installing Ruby)
* the `installation` page translation was based on a very outdated en version

### DIFF
--- a/pt/documentation/index.md
+++ b/pt/documentation/index.md
@@ -9,6 +9,20 @@ referências que serão úteis quando você sentir que gosta de programar
 em Ruby.
 {: .summary}
 
+### Instalando o Ruby
+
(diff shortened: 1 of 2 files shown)


stomar:
Merged manually.

@alanwillms Thanks for the fast response and the very good work with all the recent `pt` updates.

:clap:

Only one (tiny) suggestion: please use concise/short commit messages and branch names, since they clutter the history when they are too verbose. See the Wiki for some examples.

alanwillms:
@stomar thanks for the tip, I will keep it in mind in my next PRs :wink:

PR #1221, opened 2015-11-18
Add 'Ruby on Slack'

maclover7:


skade:
I'd avoid endorsing a commercial product in this way, especially as Slack has an upper bound on members which cannot be raised indefinitely.

hsbt:
@maclover7 Thanks for your suggestion. but I agreed with @skade 's concern.

genericlady:
I tried to register for access and nothing is happening. Should this get removed?

PR #1225, opened 2015-11-23
Enable to access Hackety Hack

ta1kt0me:
I can't access it from https://www.ruby-lang.org/en/documentation/ .


JuanitoFatas:
:+1:

PR #1227, opened 2015-11-25
Update User Group details for Lille, France

cveneziani:
- Rename group from Nord Europe Ruby User Group Francophone to Ruby Nord
- Mention friendship with Ruby Belgium community
- Update Ruby Nord website url

cc @nono :)


nono:
Thanks! It's nice to see that Ruby-Nord is active.

cveneziani:
Thanks!
We try to promote Ruby as much as we can in North of France :)

PR #1240, opened 2015-12-16
Announce for stable releases

nagachika:
Add announcement for stable releases and CVE-2015-7551.


JuanitoFatas:
ja posts missing translators.

PR #1255, opened 2015-12-26
Switch pygments to rouge

AndorChen:
On rouge's project site, it says "[Rouge's] HTML output is compatible with stylesheets designed for pygments."  And I use rouge on my personal Jekyll site since a bit long time ago, it woks very well. 

The main advantage of rouge is it's pure Ruby, no need to invoke Python to highlight.

So, I don't see any reasons to use pygments any more.

I test locally but not on Heroku, so maybe should give it a test run before switching.

What do you think about this change?

cc @hsbt @JuanitoFatas

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ ruby '2.3.0'
 
 gem 'rake',   '~> 10.0'
 gem 'jekyll', '~> 2.0'
+gem 'rouge',  '~> 1.10'
 
 gem 'unicorn'
 gem 'lanyon', '~> 0.2.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -75,6 +75,7 @@ GEM
     rb-inotify (0.9.5)
       ffi (>= 0.5.0)
     redcarpet (3.3.3)
+    rouge (1.10.1)
     safe_yaml (1.0.4)
     sass (3.4.20)
     sawyer (0.6.0)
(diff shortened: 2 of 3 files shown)


hsbt:
I agreed to use pure ruby library. It helps contributers environement easily.

@stomar How do you think about this?

JuanitoFatas:
I am +1 on this! :clap:

stomar:
I'm generally in favor, too, but before merging I would like to

* `diff` the generated html for unexpected changes
* make sure it supports all lexers we currently use, e.g. also (ba)sh
* make sure it does not (significantly) increase build time

I won't have time before January, though (and I will also not be able to participate in further discussions until then; I'm essentially offline these days).

hsbt:
I switched pygments to rouge in my hobby application. I found some incompatible change with html syntax. ex. `pre` tag in pygments changed `code` tag in rouge.

@AndorChen Can you check these changes?

AndorChen:
@hsbt I don't see this problem in a Jekyll site, maybe the reason sits in [this line](https://github.com/jekyll/jekyll/blob/v2.5.3/lib/jekyll/tags/highlight.rb#L107) of Jekyll `{% highlight %}` tag (we use this tag to highlight code blocks here): it wraps codes in a  `<pre>` tag manually.

stomar:
@AndorChen I'm about to review.
Would you please rebase your commit onto the current master (there's a conflict).

And for future PRs: note that using `master` is usually a bad idea, you should use a feature branch.

AndorChen:
@stomar I notice this after PR, sorry for that.

AndorChen:
I will close this PR and start with a new one.

Again, sorry for the mess everyone.

PR #1257, opened 2015-12-28
Update _config.yml for Ruby 2.3

sorah:
- Also moving 2.0.0 to new section called `tooold` instead of removing.
  I think leaving it for a while is better even if Ruby 2.0.0 EOL is soon, but
  it's still maintained.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -24,6 +24,18 @@ license:
 
 downloads:
   stable:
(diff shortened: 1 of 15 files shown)


makimoto:
:+1:

JuanitoFatas:
:tophat: :sparkles:

PR #1260, opened 2015-12-29
Add a space

znz:
fix wording, "Active Support" has a space in between the words, not ActiveSupport.
see http://svn.ruby-lang.org/cgi-bin/viewvc.cgi?revision=52232&view=revision


JuanitoFatas:
@znz Thanks! Merged as ab250b4 and fixes other lang posts too! :blush:

PR #1267, opened 2016-01-09
Traducir greeter.md al español

hminaya:


hminaya:
It's being handled in pull request #1270

PR #1270, opened 2016-01-09
Translate ruby examples (es)

hminaya:
Translated all 4 Ruby examples to Spanish


hminaya:
@dabit aqui esta....

dabit:
Gracias!

PR #1282, opened 2016-01-12
Vietnamese translation of 2015-12-16-ruby-2-0-0-p648-released.md

huynhquancam:


stomar:
@huynhquancam Notice the wrong lang attribute in the YAML front matter (fixed with ccde973983e8da500283bd36216148b40cf9fca9).

PR #1293, opened 2016-02-06
Fix typographical error (ES).

gfvcastro:


stomar:
@gfvcastro Welcome and thanks for your recent contributions.
Notice the commonly adopted commit message style is: `Fix typographical error (es)`

PR #1295, opened 2016-02-12
Simplify release list on downloads page (en)

stomar:
@hsbt

Stop distinguishing between "current", "previous", ... releases,
simply list all of them under "stable releases".
-> No need to come up with names; and easier to change in case there are
4 or only 2 stable branches at a certain time period, because we can simply
iterate over an array of releases defined in _config.yml.

Preview:

* Stable releases:
  * Ruby 2.3.0<br>
    sha256: ...
  * Ruby 2.2.4<br>
    sha256: ...
  * Ruby 2.1.8<br>
    sha256: ...

* In security maintenance phase (will EOL soon!):
  * Ruby 2.0.0-p648<br>
    sha256: ...

* Snapshots:
  * Stable Snapshot: ...

### DIFF
--- a/en/downloads/index.md
+++ b/en/downloads/index.md
@@ -34,22 +34,14 @@ See the [Installation][installation] page for details on building
 Ruby from source. If you have an issue compiling Ruby, consider using
 one of the third party tools mentioned above. They may help you.
 
-* **Current stable:**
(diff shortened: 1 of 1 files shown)


hsbt:
It's ok to change release name. :+1:

PR #1296, opened 2016-02-16
Switch pygments to rouge

AndorChen:
This is a new PR to replace #1255.

Repost the origin message below:

On rouge's project site, it says "[Rouge's] HTML output is compatible with stylesheets designed for pygments." And I use rouge on my personal Jekyll site since a bit long time ago, it woks very well.

The main advantage of rouge is it's pure Ruby, no need to invoke Python to highlight.

So, I don't see any reasons to use pygments any more.

I test locally but not on Heroku, so maybe should give it a test run before switching.

What do you think about this change?

cc @stomar

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ ruby ENV['CUSTOM_RUBY_VERSION'] || '2.3.0'
 
 gem 'rake',   '~> 10.0'
 gem 'jekyll', '~> 2.0'
+gem 'rouge',  '~> 1.10'
 
 gem 'unicorn'
 gem 'lanyon', '~> 0.2.0'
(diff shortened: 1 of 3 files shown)


stomar:
To do:

1. `diff` the generated html for unexpected changes
2. make sure it supports all lexers we currently use: c, diff, irb, java, perl, php, python, ruby, sh, xml
3. make sure it does not (significantly) increase build time

stomar:
Regarding 3. (speed): with `rouge` the build is even a little bit faster, at least on my machine.

stomar:
Regarding 1.: everything seems fine (only small differences in how the code is highlighted)

Only downside is that `irb` is not supported by rouge (yet?), but I think that's not crucial.

Merged as 4bf1b5e32cff455efc8e424897f4e92c5023decc; and also updated the README.

PR #1308, opened 2016-02-26
translate 2 latest news to Vietnamese

nguyenducgiang:


stomar:
@nguyenducgiang, @joneslee85 Please remember that lines should be wrapped at 80 characters max (except links).

Also, branch name need not include nick name, it appears in the merge commit anyway.

PR #1319, opened 2016-03-04
Remove space between triple backticks and language specifier [ci skip]

JuanitoFatas:


stomar:
what's the benefit?

(Personally, I like the current version more.)

JuanitoFatas:
Oh I thought without space is normal. Close then. thanks.

Issue #1321, opened 2016-03-05
Update release list on downloads page


stomar:
The `downloads` page needs an update, the way the release list is structured has changed, see e8e91bdd6783e17913f1c859983160bcb886bcbc and 2c353e400c002dfafb95ba1ac4aed27d1ffb19f5.

* [ ] bg
* [ ] de
* [ ] es (@dabit)
* [ ] fr (@chikamichi, @chatgris)
* [ ] id (@gozali)
* [ ] it (@alepore)
* [ ] ja (@makimoto)
* [ ] ko (@marocchino)
* [ ] pl (@crabonature)
* [ ] pt
* [ ] ru (@gazay)
* [ ] tr
* [ ] vi (@joneslee85)
* [ ] zh_cn (@hlcfan)
* [ ] zh_tw (@JuanitoFatas)

gmitrev:
Created a PR for the Bulgarian version - #1322

PR #1322, opened 2016-03-05
Update release list on downloads page (bg)

gmitrev:
Updates realted to issue #1321.


stomar:
That was fast! :clap:

PR #1325, opened 2016-03-11
Removed sentence of RubySpec

hsbt:
RubySpec is shutdown status and http://rubyspec.org/ is 404. It should be removed from about page.


stomar:
Or changed to https://github.com/ruby/spec ("The Ruby Spec Suite")?

stomar:
@hsbt Please go ahead and merge, I now think it's easiest to simply remove that paragraph for now.

PR #1328, opened 2016-03-22
Minor fix, spanish correction

dannyec:


hlcfan:
cool name by the way :+1:

PR #1332, opened 2016-03-25
Add atom to editor list on documentation page (ja)

snoozer05:
I found that the atom described in the English documentation page isn't in the Japanese documentation page. So I add atom to the list.


stomar:
Thanks!

PR #1334, opened 2016-03-28
Remove the feature of Ruby 1.8 (en)

snoozer05:
Because I think any newcomer don't want to get the information of old Ruby version and it can be misleading information given on the feature of current Ruby.

### DIFF
--- a/en/documentation/ruby-from-other-languages/to-ruby-from-c-and-cpp/index.md
+++ b/en/documentation/ruby-from-other-languages/to-ruby-from-c-and-cpp/index.md
@@ -78,7 +78,7 @@ Unlike C, in Ruby,...
 * There are no header files. You just define your functions (usually
   referred to as “methods”) and classes in the main source code files.
 * There’s no `#define`. Just use constants instead.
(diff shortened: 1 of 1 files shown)


stomar:
Or only remove "As of Ruby 1.8" ?

snoozer05:
The reason I remove all sentence is that I think "rather than compiled ..." is not correct and it is misleading information because actually the code is compiled to YARV bytecode instructions.

stomar:
But the main message (Ruby being an interpreted language, as opposed to C) IMO still holds. The intermediate bytecode doesn't really concern the user.

snoozer05:
I understand! so I only remove "As of Ruby 1.8".

stomar:
I'm not sure... maybe a more general version, less specific about byte/machine code?

Any third opinion?

sorah:
Just saying "no need to run compiling prior to run a code" ?

stomar:
Thanks @snoozer05 for pointing that out.

I now changed to "You don’t need to compile your code. You just run it directly." (copied from `to-ruby-from-java`, applied with 1a625d89cb3ebb86f673c5b200d8c47e93b6fff8)

PR #1330, opened 2016-03-25
translate "Ruby in Twenty Minutes" (ja)

takahashim:
There is no tutorial documents in /ja/, so I'd translated "Ruby in Twenty Minutes."

I modified the link to book guide page in the last paragraph.  This page is what I wrote last year in Rubyist Magazine.

PR #1338, opened 2016-03-31
Add 'Get Started' section to sidebar(ja)

snoozer05:
I think it is good timing to enable 'Get Started' section of sidebar of Japanese because 'Ruby in Twenty Minutes' is already translated (#1330).

So I enable and 'get_started' section and translate it.

And also remove exclamation point from translation of 'Explore' because it isn't in the original text and it is bad looking exclamation point are attached in succession.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1142,19 +1142,19 @@ locales:
           text: Ultime Notizie (RSS)
           url: /it/feeds/news.rss
     ja:
-#     get_started:
-#       text: <strong>Get Started</strong>, it's easy!
-#       try_ruby:
-#         text: Try Ruby! (in your browser)
-#         <<: *try_ruby
-#       quickstart:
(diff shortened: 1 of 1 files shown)


stomar:
Agreed!

Could you check indentation? Looks wrong, at least here on GitHub.

snoozer05:
Sorry for wrong indentation. I fixed it!

stomar:
Build fails because `quickstart` section has not enough indentation.

snoozer05:
oo... I'm sorry for lacking of confirmation. I fixed it.

stomar:
I just noticed (sorry): what about "Try Ruby", would only need translation of link text.

snoozer05:
Oh, yes. I do this!

stomar:
Thank you!

PR #1339, opened 2016-03-31
Add 'Why's (Poignant) Guide to Ruby' to Getting Started (ja)

snoozer05:


stomar:
I'm not against adding this link, but IMO rather than adding single resources, a more general sync between en/documentation and ja/documentation should be discusssed.

(Also, generally, translations should match as closely as possible, e.g. regarding link markdown, where the en version uses a numbered link reference. And please keep branch names rather short, or merge commits tend to clobber the history.)

snoozer05:
>I'm not against adding this link, but IMO rather than adding single resources, a more general sync between en/documentation and ja/documentation should be discussed.

You're right. I would like to add more links for Japanese. But I didn't find other link's Japanese resource. So I just added only 'Why's (Poignant) Guide to Ruby'.

>(Also, generally, translations should match as closely as possible, e.g. regarding link markdown, where the en version uses a numbered link reference. And please keep branch names rather short, or otherwise merge commits tend to clobber the history.)

I understand.

snoozer05:
I'll make other PR for revising Japanese getting started list.

PR #1345, opened 2016-04-03
Update book guide URL from 2015 to 2016

znz:
Released Rubyist Magazine 0053 includes new book guide.


stomar:
cc @takahashim

hsbt:
I confirmed.

PR #1369, opened 2016-04-27
Translate latest news articles into Bulgarian

gmitrev:


stomar:
Thanks. Please use `(bg)` in commit messages.

PR #1374, opened 2016-04-28
Fix syntax error in Python3

znz:


stomar:
is this also valid in older Python?

gfvcastro:
@stomar: Yes in 2.7, I do not remember in other older versions.

stomar:
Thanks!

PR #1382, opened 2016-05-07
Revise list of not overloadable operators

snoozer05:
`!`, `!=`, and `!~` are overloadable since 1.9.1 https://github.com/ruby/ruby/blob/trunk/doc/NEWS-1.9.1#L227-L228


stomar:
Thanks, good catch!

PR #1386, opened 2016-05-10
Update ruby-from-other-languages (bg)

gmitrev:
Fixes a lot of spelling and grammatical errors and updates some obsolete code examples. I also found some mistakes in the english version but I don't think it belongs here so I might open another PR with them later.


stomar:
Thanks!

PR #1388, opened 2016-05-13
Change NaCl to Ruby Association

shugo:
Please change NaCl to Ruby Association because hosting is now sponsored by Ruby Association and their sponsors.


hsbt:
👍 Thank you!

PR #1389, opened 2016-05-14
Update format for bundler 1.12.x

hsbt:
bundler 1.12.x added entry of ruby version to `Gemfile.lock` when Gemfile use `ruby` version declaration for heroku usage. I added it.

/cc @stomar


JuanitoFatas:
👍

PR #1390, opened 2016-05-14
Migrate to Jekyll 3

stomar:
Jekyll 3 comes with much less dependencies, so an upgrade would be desirable.

But there are various changes that need to be taken care of:

- plugins must be updated for changed handling of posts and collections
- layouts must be updated for changed handling of layout metadata
- various post URLs change (slightly)
- ...

I'm still in the process of investigating the changes of the rendered site.

Also, Jekyll has a bug with layout metadata that needs to be fixed before a merge (jekyll/jekyll#4897 - _fixed with jekyll 3.1.4_).


stomar:
I decided to _explicitly_ rename the posts where the URL would change, and also updated internal links and added rewrite rules (these should be removed some time later).

Then, the only remaining change of the generated `_site` is the new behavior of the `{% highlight %}` tag, that now produces highlighted code blocks that are wrapped in a `<figure>` tag instead of a `<div>` tag.

For comparison, I generated the site for the last commit before "Switch to Jekyll 3" (`_site_jekyll2`) and for the final commit of this branch (`_site_jekyll3`). Then I replaced the `figure` tags with `div` in the version generated by Jekyll 3:

```
# for zsh
$ sed -i -e 's|<figure|<div|g' -e 's|</figure>|</div>|g' _site_jekyll3/**/*.html
$ sed -i -e 's|\&lt;figure|\&lt;div|g' -e 's|\&lt;/figure\&gt;|\&lt;/div\&gt;|g' _site_jekyll3/**/news.rss
```
and did a `diff` of the two sites, ignoring whitespace changes:
```
$ diff -r -Bb _site_jekyll2 _site_jekyll3
```
The output shows only the expected difference in the generation time (in `admin/index.html`).

stomar:
... and deployed :fireworks:

PR #1226, opened 2015-11-24
to-ruby-from-python functional programming

StephenLujan:
Added one bullet point of similarities and two of differences, regarding functional programming, in to-ruby-from-python/index.md

PR #1394, opened 2016-05-19
Translate to-ruby-from-python (zh)

hlcfan:
Update per https://github.com/ruby/www.ruby-lang.org/pull/1226

@JuanitoFatas Please review.


JuanitoFatas:
Awesome, thanks 👍

PR #1395, opened 2016-05-19
Change NaCl to Ruby Association (zh)

hlcfan:
Update per https://github.com/ruby/www.ruby-lang.org/pull/1388


JuanitoFatas:
Thanks! 

However, commit message should follow [this guideline](https://github.com/ruby/www.ruby-lang.org/wiki#commits-and-commit-messages) 😅

stomar:
Yes, only plain ASCII, please; not only in the first commit msg line.

PR #1398, opened 2016-05-19
Markup the summary section on root pages(ja)

snoozer05:
I think it is better to fit the style of Japanese translation to other languages.
So I add the summary section (and markup it) to root pages of Japanese translation.


stomar:
Thank you! (Merged manually as 0fc847f3a6e2ce0f57129b49bd66db7c85fb88f4.)

PR #1404, opened 2016-05-30
Add Ruby file extension to docs

Oisins:
Added "Ruby scripts have the file extension `.rb`" to the switching from python page, as it took me a long time to initialy find this info.


stomar:
Thanks for pointing this out.

However, regarding the proposed changed I have objections:

* this is not specific to `to-ruby-from-python`
* the statement over-simplifies (e.g. I have many Ruby scripts without any extension)

I'll think about a better place for this info later.

(And please distinguish between PR description and commit message, which is too verbose and also doesn't follow our conventions.)

PR #1408, opened 2016-06-08
Update instructions for dnf package management

arjkb:
CentOS, Fedora, and RHEL now use the dnf package manager instead of the older yum package manager. Updating the site to reflect this change.


stomar:
Please confirm that CentOS and RHEL already use dnf by default. From a (superficial) research I got the impression that the default package manager still is yum.

hsbt:
`dnf` is only enabled Fedora 22 or later.

PR #1418, opened 2016-06-27
Add title and charset on index.html page

spk:
This fixes 2 errors

* Error: The character encoding was not declared.
* Error: Element head is missing a required instance of child element title.

Cheers


stomar:
Thanks!

PR #1420, opened 2016-06-29
Validate with libtidy on travis

spk:
Use libtidy https://github.com/htacg/tidy-html5
for complete validation of the website, just excluding `examples` directory (inserted in javascript).
Also update validate-website gem.

Time ~ 40sec for `check:markup` cf https://travis-ci.org/ruby/www.ruby-lang.org/builds/501125031#L582

Cheers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
+---
 language: ruby
 sudo: false
 rvm: 2.5.1
 cache:
   bundler: true
-script: bundle exec rake ci
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then wget https://github.com/htacg/tidy-html5/releases/download/5.4.0/tidy-5.4.0-64bit.deb; sudo apt-get remove '^libtidy.*';  sudo dpkg -i tidy-5.4.0-64bit.deb; fi
+script:
+  - bundle exec rake ci
+  - bundle exec rake check:markup
 # Notifications, used by our Gitter channel.
 notifications:
   webhooks:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
----
 language: ruby
 sudo: false
 rvm: 2.5.1
--- a/Gemfile
+++ b/Gemfile
@@ -18,5 +18,5 @@ end
 group :development do
   gem 'minitest'
   gem 'spidr', '~> 0.6'
-  gem 'validate-website', '~> 1.6'
+  gem 'validate-website', '~> 1.9'
 end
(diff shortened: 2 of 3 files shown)


stomar:
HTML validation on Travis has been discussed before, I think.

IMO it's a bad idea to check every build:

* it's really not necessary since the HTML layouts very rarely change; most commits only add/change markdown, which usually does not introduce malformed HTML
* it takes way too much time (the build of this PR took **8 min 22 s**); note that there already have been complaints about the current 2...3 min

I rather think that occasional validations should be sufficient.

spk:
Hi, I've updated the PR using libtidy (around 40sec to check all files) and fixes the errors hope we could reconsider including the changes, thanks

.travis.yml
```diff
@@ -1,9 +1,14 @@
+---
```
stomar: unrelated change

```suggestion
```

stomar:
It‘s a little confusing: this is supposed to target CI, but commit message says „locally“.

.travis.yml
spk: Its yaml headers recommended by linters should i remove it ?

spk:
> It‘s a little confusing: this is supposed to target CI, but commit message says „locally“.

you are right its confusing i will remove this

.travis.yml
spk: Removed in https://github.com/ruby/www.ruby-lang.org/pull/1420/commits/fea533eb88543172cf6acc8842b2efebd7f96db0

stomar:
After a closer look I feel I have to close this PR, since it doesn‘t make any sense to run on CI in this form, and would even be harmful.

The problem is that the validation task does not check the current build but the already deployed site. This would mean that a PR introducing errors would **not** fail on CI, but all subsequent PRs **would**, regardless of whether they introduce errors or not.

stomar:
As to the general question whether this check should (in a fixed version) be added to CI: 40 seconds is not too bad, but personally I’m still undecided and would like to have some other opinions.

@hsbt

spk:
> The problem is that the validation task does not check the current build but the already deployed site. This would mean that a PR introducing errors would not fail on CI, but all subsequent PRs would, regardless of whether they introduce errors or not.

This is **not** correct this test the current branch since it check the build-ed `_site` generated on CI. cf https://github.com/ruby/www.ruby-lang.org/blob/master/lib/markup_checker.rb#L7

spk:
> As to the general question whether this check should (in a fixed version) be added to CI: 40 seconds is not too bad, but personally I’m still undecided and would like to have some other opinions.
> 
> @hsbt

If this something the team should talk about could you please reopen the PR @stomar please ?

spk:
Reopened in https://github.com/ruby/www.ruby-lang.org/pull/1990 to have team feedback

stomar:
Then please remind me what is the meaning of the `—site https://www.ruby-lang.org/` option?

PR #1421, opened 2016-06-30
<meta charset> must be before <title>

znz:
see https://code.google.com/archive/p/doctype-mirror/wikis/MetaCharsetAttribute.wiki


stomar:
Thanks!

PR #1422, opened 2016-06-30
Fix channel in link

znz:


stomar:
Does the other channel not exist anymore?

znz:
`#ruby:*.jp` on IRCnet exists even now.
But there are few messages in recent years on `#ruby:*.jp`.

PR #1425, opened 2016-07-11
Fix hint on evaluation strategy of methods

neumanrq:
After investigating this question, I'm convinced that Ruby does not use *call-by-reference*, as stated in the section "To Ruby From C and C++" . It is using *call-by-value* as evaluation strategy for methods in canonical terminology.  

See [this blogpost](http://robertheaton.com/2014/07/22/is-ruby-pass-by-reference-or-pass-by-value) or [this stackoverflow discussion](http://stackoverflow.com/questions/1872110/is-ruby-pass-by-reference-or-by-value) for deeper discussions on that question.

Obviously this information causes some confusion. For example, [the wikipedia article on evaluation strategies](https://en.wikipedia.org/wiki/Evaluation_strategy) referenced the ruby-lang documentation as a proof that Ruby does use *call-by-reference*:
> whereas in the Ruby community, they say that Ruby is call-by-reference

That's why I'd suggest changing the information contained on the mentioned section. 

### Translations

I added the fix for German and English version, but since I'm not able to serve verified suggestions for the other languages, maybe the lang maintainers could assist?

- [ ] bg @mytrile
- [ ] fr @nono 
- [ ] id @gozali
- [ ] it @kennyadsl
- [ ] ja @take 
- [ ] zh_cn @AndorChen
- [ ] zh_tw @JuanitoFatas
- [ ] pt_br @fuadsaud
- [ ] pl @crabonature
- [x] ru @gazay
- [ ] vi @joneslee85

Thanks!

### DIFF
--- a/en/documentation/ruby-from-other-languages/to-ruby-from-c-and-cpp/index.md
+++ b/en/documentation/ruby-from-other-languages/to-ruby-from-c-and-cpp/index.md
@@ -81,8 +81,8 @@ Unlike C, in Ruby,...
 * There’s no `#define`. Just use constants instead.
 * All variables live on the heap. Further, you don’t need to free them
   yourself—the garbage collector takes care of that.
-* Arguments to methods (i.e. functions) are passed by reference, not by
-  value.
+* Arguments to methods (i.e. functions) are passed by value, whereas the
+  values are always references
 * It’s `require 'foo'` instead of `#include <foo>` or `#include "foo"`.
 * You cannot drop down to assembly.
 * There’s no semicolons ending lines.
--- a/en/documentation/ruby-from-other-languages/to-ruby-from-c-and-cpp/index.md
+++ b/en/documentation/ruby-from-other-languages/to-ruby-from-c-and-cpp/index.md
@@ -81,7 +81,7 @@ Unlike C, in Ruby,...
 * There’s no `#define`. Just use constants instead.
 * All variables live on the heap. Further, you don’t need to free them
   yourself—the garbage collector takes care of that.
-* Arguments to methods (i.e. functions) are passed by value, whereas the
+* Arguments to methods (i.e. functions) are passed by value, where the
   values are always references
 * It’s `require 'foo'` instead of `#include <foo>` or `#include "foo"`.
 * You cannot drop down to assembly.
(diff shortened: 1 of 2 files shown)


stomar:
Note that the cited blog post has syntactically wrong code examples, which does not inspire confidence...

Can someone provide a more "official" view point (ruby core)?

neumanrq:
Maybe [this answer based of the mentioned stackoverflow discussion](http://stackoverflow.com/a/18069011/2159942) does contribute to the trustworthiness, because it quotes the O'Reilly book "The Ruby Programming Language".

stomar:
Great, that's exactly the kind of reference I had in mind :)

Here the quote from Flanagan/Matsumoto as given on stackoverflow:

> [from 3.8.1: Object References]
>
> When you pass an object to a method in Ruby, it is an object reference that is passed to the method. It is not the object itself, and it is not a reference to the reference to the object. Another way to say this is that method arguments are passed by value rather than by reference, but that the values passed are object references.
>
> Because object references are passed to methods, methods can use those references to modify the underlying object. These modifications are then visible when the method returns.

en/documentation/ruby-from-other-languages/to-ruby-from-c-and-cpp/index.md
```diff
@@ -81,8 +81,8 @@ Unlike C, in Ruby,...
 * There’s no `#define`. Just use constants instead.
 * All variables live on the heap. Further, you don’t need to free them
   yourself—the garbage collector takes care of that.
-* Arguments to methods (i.e. functions) are passed by reference, not by
-  value.
+* Arguments to methods (i.e. functions) are passed by value, whereas the
```
stomar: where ... object references

stomar:
@neumanrq wrote
> Obviously this information causes some confusion. For example, the wikipedia article on evaluation strategies referenced the ruby-lang documentation as a proof that Ruby does use call-by-reference

I think you misunderstood the Wikipedia article. It clearly states that Ruby does not use "call by reference" but what they term "call by sharing" (i.e. call by value where the value is a reference); the paragraph you have cited only makes a statement about the used terminology:

> However, the term "call by sharing" is not in common use; the terminology is inconsistent across different sources. For example, in the Java community, they say that Java is call by value, whereas in the Ruby community, they say that Ruby is call-by-reference, even though the two languages exhibit the same semantics.

I have no idea what the predominantly used terminology in the community is, though.

neumanrq:
> I think you misunderstood the Wikipedia article. It clearly states that Ruby does not use "call by reference" but what they term "call by sharing" (i.e. call by value where the value is a reference); the paragraph you have cited only makes a statement about the used terminology:

Hm, yes, you're right. I fully agree to the wikipedia argumentation on *call-by-sharing*. But the issue here remains: I still disagree to the statement
> whereas in the Ruby community, they say that Ruby is call-by-reference

This statement was underlined by quoting ruby-lang documentation. That was the reason to make this proposal. 

I fixed the wording due to your suggestion

PR #1427, opened 2016-07-12
Fix hint on evaluation strategy of methods (it)

kennyadsl:
@ruby/www-ruby-lang-org-i18n-it please review, ref #1425


alepore:
👍

PR #1431, opened 2016-07-25
Update cities.md

lucascaton:


stomar:
This is no typo. In case of differences between pt-PT and pt-BR we use the Portuguese version (same as Wikipedia does), which is Amesterdão and not Amsterdã, AFAIK.

lucascaton:
@stomar you're right. Sorry, I didn't realised it was `pt-PT` and not `pt-BR `.

PR #1432, opened 2016-07-25
Replace 2.1.9 with 2.1.10 in stable versions list

lucascaton:


hsbt:
@lucascaton Thank you!

stomar:
Should 2.1.10 be "advertised" as stable release?

From the release notes:

> This release is not intended for production use, but for compatibility tests with two-digit version numbers. You don’t have to replace Ruby 2.1.9 by 2.1.10 in normal use.

lucascaton:
@stomar I think it should as it doesn't hurt, otherwise it may be confusing.

PR #1436, opened 2016-08-01
Use a newer Rails version in "gem" command example

hlcfan:


stomar:
Since the example is meant to illustrate how to install a specific version of a gem (and _not_ the latest, which is installed by default), using the latest version would not be very illustrating.
(`gem install rails` and `gem install rails --version 5.0` currently have the same effect.)

_IMO it's not even necessary to keep the version in the example code up-to-date; this is not an installation instruction for Rails but a short overview of the `gem` command. However, I do not mind using a more current one._

I propose to wait some time before merging... (at least until 5.1 is released).

_Note: "squash and merge" commit with new commit message `Use a newer Rails version in code example` or similar_

stomar:
Rails 5.1 has been released.

PR #1438, opened 2016-08-04
Removed not working sites, updated links numbers

piton4eg:


hsbt:
👍  Thank you!

stomar:
Note that link numbers are not and should not be updated. This makes diffs between commits and also between `en` and translated pages **much** more easier to read. See commit 582e4de62b5809c94a1ebe7345d8a2078a2c9393 as demonstration.

(I will add a corresponding remark to the Wiki.)

PR #1439, opened 2016-08-05
Removed not working links (zh_cn)

hlcfan:


stomar:
See my recent commits:

* link numbers should be kept unchanged
* this kind of change can and should be done in bulk for **all** translations, not individually

PR #1446, opened 2016-09-01
Update about page to reflect Windows OS'

ashleydavies:
The Windows OS' listed were 10 years out of date, swapped the order with Mac OS X so they would fit on a line in the file and added Vista/7/8/10.


ashleydavies:
Just realised I should update the other languages. Will issue a new pull request once I finish.

Issue #1447, opened 2016-09-01
Supported versions of Windows listed are 10 years out of date


ashleydavies:
In the index of 'about,' the listed versions of Windows go up to XP, which was now released 15 years ago.

I propose this change:

Ruby is highly portable: it is developed mostly on GNU/Linux, but works on many types of UNIX, Mac OS X, Windows 95/98/Me/NT/2000/XP, DOS, BeOS, OS/2, etc

Ruby is highly portable: it is developed mostly on GNU/Linux, but works on many types of UNIX, Mac OS X, Windows since Windows 95, DOS, BeOS, OS/2, etc

I would normally issue a pull request with this change myself however I suspect it may be preferred if multiple languages are updated at once and I don't trust myself to get the translations sounding correct.

riseshia:
I think it's good!
how do you want to handle this?

ashleydavies:
I'm happy to issue a pull request for English, German and Spanish. Will do those tonight. I assume it's OK to not update every language at once.

stomar:
@ashleydavies Thanks for pointing this out. It's really not necessary to point out any specific version (it's also not done for Mac OS X).

PR #1448, opened 2016-09-02
Update supported version of windows (ko)

riseshia:
#1447 
@ruby/www-ruby-lang-org-i18n-ko
Please review this!


marocchino:
Shouldn't we wait for the english version first?

riseshia:
I thought that's enough to do... ;_;
anyway, ok, I will close this and let's wait english version.

Issue #1405, opened 2016-06-02
Use of “we” in news posts


Quintus:
Hi everyone,

while translating the [ConFoo Vancouver 2016 post](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2016-05-16-confoo-cfp.md) I thought about these lines (emphasis by me):

> **We** will cover travel and hotel for the speakers who require it. […] Visit **our** site to learn more. […] **We** are eagerly expecting your proposals!

This post on ruby-lang.org invokes the association that this “we” refers to the Ruby core developers, or the organisation running ruby-lang.org (RubyCentral?). I think that the use of “we” and related pronouns is best reserved for referring to the Ruby core developers themselves as ruby-lang.org is the official website of the Ruby programming language, which is best represented by its core developer team (as in “we are happy to announce Ruby 2.3.0”). Any other use of “we” is misguiding and should best be avoided.

In this particular case, I would rather have suggested to use something like (differences emphasized):

> **ConFoo** will cover travel and hotel for the speakers who require it. […] Visit **their** website to learn more. […] **They** are eagerly expecting your proposals!

What do you think?

Valete,
Marvin

hlcfan:
👍 Make sense

stomar:
@Quintus

I completely agree, "we" should be reserved for ruby core or the ruby-lang.org team.

Unless... what about posts signed with sth like "the ConFoo team"; still too confusing (you would have to look at the very end of the post) or acceptable?

Quintus:
Am Sun, 05 Jun 2016 13:07:33 -0700
schrieb Marcus Stollsteimer <notifications@github.com>:

> Unless... what about posts signed with sth like "the ConFoo team";
> still too confusing (you would have to look at the very end of the
> post) or acceptable?

If moved to top and written like this:

“This is a guest post by the ConFoo team.”

I would be fine with it. Simply not using “we” however seems more
reasonable to me.

Vale,
Quintus

stomar:
Agreed, simplest version is avoiding "we" for third-party posts.

jonjensen:
It looks like this issue was taken care of, but never closed and should be closed.

JuanitoFatas:
Thanks @jonjensen <3

PR #1449, opened 2016-09-02
Fix ConFoo post

stomar:
* remove banner from first paragraph: this paragraph is shown as excerpt on the main page and archive pages and should not have banners in it
* avoid use of "we" [**only fixed in `en` version**]: "we" suggests ruby-core or ruby-lang.org, so it should be avoided in third-party posts (see #1405)
* fix typo for `id` (s/Confoo/ConFoo/)


afilina:
Good points, thanks! I'll refer to this next time I post an announcement.

hsbt:
👍  Thanks always.

PR #1450, opened 2016-09-02
Avoid use of "we" in ConFoo post (ko)

yous:

See #1405. Other changes made in f21271e5ec15966bba0db045a29d5d96e602091a are already context-free in Korean.


riseshia:
LGTM!

PR #1458, opened 2016-09-18
Fix "inscrire" link on 2007-10-20-euruko-2007

spk:
Generated as https://www.ruby-lang.org//fr/news/2007/10/20/euruko-2007/:http://www.approximity.com/cgi-bin/europeRuby/tiki.cgi?c=v&p=Registration2007
On https://www.ruby-lang.org//fr/news/2007/10/20/euruko-2007/


JuanitoFatas:
Thanks!

PR #1478, opened 2016-10-14
Corrected a sentence and a broken link

neistuff:
About the french books link: it now leads to a website in japanese for some unknown reason. I've searched on Google for a list of books available to learn Ruby, and I found this list on developpez.com: http://ruby.developpez.com/livres/ which is a useful resource to learn programming in a lot of languages, so I thought this link might be useful. There are some duplicates with the english list, though. I cannot choose whether I put it in or not, so it's up to you. I removed it for now.

(Also it's my first pull request ever (yaaaaay!), so I'm not sure if I'm doing it properly, but here it is anyway. Feel free to give me some advice!)


stomar:
Your first PR is merged :fireworks:

Thanks!

I amended the commit message; note that PR description and commit message are different things. The latter should describe the actual change and, possibly, reasons for it -- if necessary -- in a concise way. Your changes were so small that a commit message body isn't really necessary. For the PR description it's perfectly ok to provide more thoughts, alternatives, etc.

I'm not sure about the possible new book list, yet.

PR #1485, opened 2016-10-28
Translate Fukuoka Ruby Award 2017 news (id)

meisyal:
Just translated the latest news about Fukuoka Ruby Award 2017 into bahasa Indonesia. This translation needs a review. cc: @gozali and @stomar.

Thank you.


gozali:
+1

PR #1489, opened 2016-11-11
Add link to ruby issues

riseshia:
this pr have changes like this:

- Add link from feature text to redmine.

I wonder It just missed or not, anyway, It's better to have the link I think. :)


hsbt:
👍  Helpful. Thank you.

PR #1493, opened 2016-11-15
Fix display of the download section of 2.4.0-preview3 release post (en, ja)

JuanitoFatas:
This PR fixes 2.4.0-preview3 posts' download section for en and ja.

Instead of showing 

screenshot 2016-11-15 21 01 54

Show in code blocks instead as [other](https://www.ruby-lang.org/en/news/2016/11/15/ruby-2-3-2-released/) [release](https://www.ruby-lang.org/en/news/2016/11/15/ruby-2-2-6-released/) posts do it:

screenshot 2016-11-15 21 01 46


yous:
I think then download section for ko should be changed also.

https://github.com/ruby/www.ruby-lang.org/blob/master/ko/news/_posts/2016-11-09-ruby-2-4-0-preview3-released.md

JuanitoFatas:
> I think then download section for ko should be changed also.
> 
> https://github.com/ruby/www.ruby-lang.org/blob/master/ko/news/_posts/2016-11-09-ruby-2-4-0-preview3-released.md

Thank you! Fixed in bfb0bfa.

PR #1506, opened 2016-11-24
Translate November ’16 posts (pt)

filiperocha:


filiperocha:
meanwhile found a typo.

PR #1522, opened 2016-12-04
Add information on Ruby releases and branches

stomar:
This PR adds a `releases` and a `branches` page with information on all Ruby releases resp. on the maintenance status of the various Ruby branches.

The pages are not yet linked to from the existing content; IMO the most appropriate section would be `downloads`.

Related to #1521.

@hsbt, @makimoto, @sorah, @JuanitoFatas


stomar:
Preview of the `releases` page:



Preview of the `branches` page:

hsbt:
I prefer to publish this page to ruby users.

@unak @nagachika @nurse 

Can you update  `_data/releases.yml` and ` _data/branches.yml` when you release new version of ruby?

JuanitoFatas:
@stomar Should we explain our version numbers and defined each stages (normal, security, preview, eol). Similar to [Rails' maintenance policy](http://guides.rubyonrails.org/maintenance_policy.html)?

nagachika:
I'm okey to update `_data/releases.yml` when I make stable releases.
I have a question. The `downloads` section in `_config.yml` will be deprecated instead of them?

And I have an opinion. Please sort the entries of `_data/releases.yml` by release date in `releases/index.md `. We (@unak and I) sometimes experienced the conflicts in `_config.yml` when we made pull requests for stable release announcements. We can edit `_data/releases.yml` without conflict if it could contains entries ordered by branches.

stomar:
@JuanitoFatas I added at least basic explanations of the various stages.

@nagachika I reordered the entries in the file by version number.

Regarding `_config.yml`: I already prepared moving the `downloads` section into a file in the `_data` directory and will merge soon (I think it's good to separate this data from the other stuff in the config file, mostly locales). But that's only a minor "technical" change.
I'm not sure yet how the downloads data could be directly integrated with the releases data in a simple way. The logic of picking the proper releases to display on the downloads page cannot be too complex, because we have to use the tags and filters that Liquid offers. There might come up a solution eventually...

So, for some time at least there would be some duplication, with separate downloads data.
However, IMO it's not mandatory that the releases page is updated the second a new release is available, since the new releases are listed on the downloads page anyway. Any maintainer of www.ruby-lang.org could add the data later.

@hsbt If you give a `+1` I will merge the PR as soon as I have checked all the links.

stomar:
@hsbt @unak @nagachika @nurse

The downloads data (urls, checksums) has been moved from `_config.yml` to `_data/downloads.yml`, but besides the location nothing has changed. I most probably will not think about incorporating the downloads data with the other release data (`_data/releases.yml`) before next year.

I tried to prevent future merge conflicts by adding newlines (6d740616f03370a3ffd6e0dbf188d7b091df23fc, the only vital newline is between the 2.2 and 2.3 information). I think git uses a context of one line, so this should help.

PR #1525, opened 2016-12-12
Ruby 2.4.0-rc1 Released

nurse:


en/news/_posts/2016-12-12-ruby-2-4-0-rc1-released.md
```diff
@@ -0,0 +1,134 @@
+---
+layout: news_post
+title: "Ruby 2.4.0-rc1 Released"
+author: "naruse"
+translator:
+date: 2016-12-12 00:00:00 +0900
```
stomar: @naruse Since we use UTC, which in this case is 2016-12-**11** 15:00:00 UTC, this is displayed as posted on `11 Dec 2016`, also the URL is `/en/news/2016/12/11/ruby-2-4-0-rc1-released/`.

File name, time stamp, and URL should match IMO. Unless you care for the specific time you have used we could change to e.g. 2016-12-12 12:00:00 +0000 or 2016-12-12 12:00:00 +0900 (this would be the same date both for UTC and JST, namely 12 Dec).

I would fix it, when you tell me which time stamp I should use.

stomar: Considering the current local time in Japan I was so bold to change without waiting for feedback.

I used `2016-12-12 18:00:00 +0900`, similar to various other releases, assuming the intention of having 2012-12-12 as date.

PR #1546, opened 2017-01-04
Translate Ruby 2.4.0-rc1 (es)

FerPerales:


stomar:
Thanks. How about the final 2.4.0 post? Seems more important than rc...

cc @dabit

FerPerales:
@stomar Tackle that one on a different PR: https://github.com/ruby/www.ruby-lang.org/pull/1547 cc @dabit

stomar:
Yep, I've noticed too late.

stomar:
Small fixes in 14d0da19e3982a3e6bfd2a4e97c3c85f892d521f.

Please wrap at max. 80 characters and keep same paragraphs as `en` version.

PR #1552, opened 2017-03-15
Added Nanobox under Platforms

sdomino:
Added a new section titled "Platforms" for installation options that have built-in package management, installation, and version management.

Added Nanobox as an installation option under Platform


stomar:
I'm sorry, but this is not the right place for commercial services.

PR #1553, opened 2017-03-20
Feature/nanobox

sdomino:
I updated the PR to link to our open-source github project rather than our site (which was what I meant to do the first time).

I also removed much of the description to make it more concise, and less about Nanobox and more about the benefits to Ruby developers.

I don't know if this will make a difference or not, but either way, thank you for your time to consider the pull requests.

edit: In response to #1552

### DIFF
--- a/en/documentation/installation/index.md
+++ b/en/documentation/installation/index.md
@@ -45,6 +45,8 @@ for different needs and platforms.
   * [rbenv](#rbenv)
   * [RVM](#rvm)
   * [uru](#uru)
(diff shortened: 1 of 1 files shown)


stomar:
My objection from the first PR still holds. It's not about wording or about where the actual link points to, etc.; this page is not intended as a place for promoting commercial development/deployment services.

cc @hsbt

sdomino:
Understandable. Thank you.

hsbt:
@stomar I agreed your opinion.

PR #1563, opened 2017-03-28
Fix wrong url

kakutani:
the latest release is 0055.

via https://twitter.com/tmtms/status/846549438164230145


kakutani:
dup of #1562

PR #1581, opened 2017-04-11
Mention about an API incompatibility of Ruby 2.3.4

unak:
Please review my poor English.

PR #1582, opened 2017-04-12
Mention about an API incompatibility of Ruby 2.3.4 (zh_cn)

hlcfan:
Following of https://github.com/ruby/www.ruby-lang.org/pull/1581


hlcfan:
@AndorChen Thank you, please review again.

Issue #1524, opened 2016-12-04
Update success stories


RTJ:
I think, would be nice to update info on:
https://www.ruby-lang.org/en/documentation/success-stories/

Maybe it will be good to reject **Blue Sequence** from Web Applications (the current provided link
http://www.bluefountain.com/supply-chain-management/ points to **Django debug error page**, I think that's ridiculous.)

And I think it would be nice to add links like GitHub, Shopify, Twitch, Kickstarter ? (
Or link to top ruby / rails apps or projects,

The information is very outdated ... and the **current information inspires to NOT use ruby** ...

matkoniecz:
https://github.com/ruby/www.ruby-lang.org/blob/64321ebc4abb1dbd8b1712b3a73418f2ae389970/en/documentation/success-stories/index.md is the affected file

matkoniecz:
It appears to be duplicate of  #732

siaw23:
+1

hsbt:
The Blue Sequence link was removed in #1593.

PR #1593, opened 2017-06-05
remove dead link - work on #1524

matkoniecz:


JuanitoFatas:
Thank you!

stomar:
@matheuslao @JuanitoFatas 

for future PR's: please do not change references for unaffected links when removing a reference, it makes reading the diff much more difficult and has no benefit, since the reference numbers do not show up on the rendered page.

See https://github.com/ruby/www.ruby-lang.org/wiki#markdown-link-references-.

PR #1600, opened 2017-06-21
Update Homebrew section on installation page (es)

esparta:
El Capitan and macOS Sierra were not there. Also change 2.0.0 to use 2.0 as used on english documentation


stomar:
Thanks!

PR #1605, opened 2017-07-03
Adding To Ruby from Python (es)

esparta:


esparta:
I just amend the commit since it had some typos (now fixed).

JuanitoFatas:
Could someone help verify this? @ruby/www-ruby-lang-org-i18n-es Thanks! 🇪🇸

PR #1610, opened 2017-07-16
Reword the Installing Ruby section in documentation

colby-swandale:
Hi! 👋 

This PR is the first in a few issues i have with the wording on the ruby-lang.org website which i would like to propose. This first PR aims to reword the documentation section on "Installing Ruby" (also including the page subtitle) by making it more clear and to the point on how to install Ruby.

Let me know what you think.

### DIFF
--- a/en/documentation/index.md
+++ b/en/documentation/index.md
@@ -5,23 +5,26 @@ lang: en
 ---
 
 Here you will find pointers to manuals, tutorials and references that
-will come in handy when you feel like coding in Ruby.
+will come in handy.
 {: .summary}
 
 ### Installing Ruby
 
-Unless you only want to try Ruby in the browser (see the links below)
-you need to have Ruby installed on your computer.
-You can check whether Ruby already is available by opening a terminal
-and typing
+Ruby is readily avaliable on all major platforms including Linux, Windows
+and MacOS. Detailed instructions on installing Ruby can be found in the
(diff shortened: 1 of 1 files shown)


stomar:
Thanks for the PR, I'll look into it.

en/documentation/index.md
```diff
@@ -5,23 +5,26 @@ lang: en
...
+will come in handy.
 {: .summary}
 
 ### Installing Ruby
 
-Unless you only want to try Ruby in the browser (see the links below)
-you need to have Ruby installed on your computer.
-You can check whether Ruby already is available by opening a terminal
-and typing
+Ruby is readily avaliable on all major platforms including Linux, Windows
```
tjouan: I think there is a small typo on "available".

en/documentation/index.md
```diff
@@ -5,23 +5,26 @@ lang: en
...
+and MacOS. Detailed instructions on installing Ruby can be found in the
+[installation documentation](installation/).
+
+To check if Ruby has been installed succesfully, open a new terminal
+session and enter:
```
tjouan: I'm not sure, but I guess the idea of previous text was that Ruby may
already be available, so it would not necessarily need to be
installed.
Also, just my opinion but I find that "terminal" and "session" might
be slightly confusing and not always applicable, we just need to
"execute this command".

PR #1612, opened 2017-07-19
remove gmane graph from the about page 

colby-swandale:
The Gmane service is no longer available and the graph images on the about page are now showing as broken. This PR just removes the gmane images where used.

screen shot 2017-07-19 at 10 01 52 pm

Thanks!


hsbt:
Good catch. Thanks.

PR #1611, opened 2017-07-18
bump ruby version in apt installation docs

colby-swandale:
This fixes #1609 by bumping the version of ruby that `full-ruby` is currently installs on the most recent release of Ubuntu.

PR #1614, opened 2017-07-19
Update the most recent release of Ruby for Debian and Ubuntu

meisyal:
As mentioned on #1611, the most recent release of Ruby for Debian and Ubuntu is 2.3.1. These changes include other languages. I saw outdated installation page on some languages (`de`, `it`, `ja`, `pl`, and `zh_tw`). Beside that, there is **no installation page** for `tr` language. What do you think?
@stomar


hsbt:
Thanks.

We should give up to update outdated locales.

meisyal:
You're welcome.
OK. I hope the contributors for those languages are aware.

znz:
How about using variables like downloads page?

meisyal:
Ah, it's a good idea @znz. Using that method, we don't need to update each page when new version of Ruby is available. Should I work on it?

znz:
@meisyal Yes, please.

meisyal:
OK. Where should I put variables that contain Ruby version?
There are some `.yml` files.

stomar:
Not sure whether it's worth while to keep the version information for specific distributions, or whether we should simply drop them.

(Regarding `tr`: it's unmaintained for a long time.)

znz:
Debian stretch (stable)'s [ruby](https://packages.debian.org/search?keywords=ruby) is 2.3.3.
And Ubuntu 16.04.3 LTS's ruby is 2.3.1.

```
% lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.3 LTS
Release:        16.04
Codename:       xenial
% /usr/bin/ruby -v
ruby 2.3.1p112 (2016-04-26) [x86_64-linux-gnu]
```

So it may be better to split versions to Debian and Ubuntu. Or it may be better to remove mention to version like yum and others.

stomar:
Let's drop them. When we start splitting information for Debian and Ubuntu we should as well start to list Ruby versions for the various distribution versions, like Ubuntu 17.10, 17.04, ..., Debian Jessie, ... It would need incessant maintenance.

PR #1620, opened 2017-08-08
update ruby version provides by ruby-full package

startaper:


hsbt:
ref. https://github.com/ruby/www.ruby-lang.org/pull/1614

hsbt:
Thanks.

PR #1637, opened 2017-09-14
Add 2.4.2 to Releases page

patientdev:
It looks like this is breaking the official Docker builds for Ruby (https://github.com/docker-library/ruby/pull/152#issuecomment-329588704).

### DIFF
--- a/_data/releases.yml
+++ b/_data/releases.yml
@@ -1,5 +1,8 @@
 # 2.4 series
 
+- version: 2.4.2
+  date: 2017-09-14
+  post: /en/news/2017/09/14/ruby-2-4-2-released/
 - version: 2.4.1
   date: 2017-03-22
   post: /en/news/2017/03/22/ruby-2-4-1-released/


hsbt:
Thanks!

stomar:
Interesting... didn‘t know the releases page is used in this way.

So ideally the page should be updated together with the publication of the release announcements.

patientdev:
Yep. Docker is [also considering](https://github.com/docker-library/ruby/commit/2925aed5fe286a35db19fcc770379234a425cc3c) (see added TODO) using  [`downloads.yml`](https://github.com/ruby/www.ruby-lang.org/blob/master/_data/downloads.yml).

A simple API to retrieve Ruby versions info would be worthwhile, I think, if using `downloads.yml` isn't tenable. The [RubyGems API](http://guides.rubygems.org/rubygems-org-api/) might allow adding the information, or have an API hosted by ruby-lang.org

stomar:
Anyone using the yml files should be aware that they were never considered an official public "API" to the respective data, but only a means to generate the web pages. They might change anytime without further notice (which doesn't mean that I expect them to change constantly, of course).

They are also currently not part of the "standard procedure" of the Ruby core team for releases, afaik; some branch maintainers remember to update them at release time, some don't, and the information is then added later by contributors to ruby-lang.org.

patientdev:
I'd like to recommend that ruby-lang.org provide an easily-parseable document containing information found in the `data.yml` and `releases.yml` file; specifically version number, release date, and release type (e.g., preview, stable, eol, etc.).

PR #1660, opened 2017-10-10
Update Multiple vulnerabilities in RubyGems post (id)

meisyal:
This update contains:
- Mention about upgrading Rubies as mentioned on commit 021a7379d46dac97b5716f9cf5841866c1318d47.
- Fix date format from `DD-MM-YYYY` to `YYYY-MM-DD` because previous posts use this format (consistency).


JuanitoFatas:
Thanks as always 👍

PR #1667, opened 2017-10-14
Run test on Travis CI

mtsmfm:
Currently, CI is passed even though test task is failed


stomar:
That was by intention; the test tasks were meant more as a kind of linter, and were run only occasionally (manually) to not slow down CI. But anyway, I don't object to running those always if nobody minds the extra time (not sure how much it is, actually).

mtsmfm:
@stomar 

> That was by intention

Sorry, I didn't know about that 💦 

According to the following snnipet, CI got 40 seconds slower.

```ruby
[56] pry(main)> Travis::Repository.find('ruby/www.ruby-lang.org').builds.lazy.select(&:green?).select {|b| b.branch_info == 'master' && b.number.to_i < 4287 }.take(10).sum(&:duration) / 10.0
=> 220.2
[57] pry(main)> Travis::Repository.find('ruby/www.ruby-lang.org').builds.lazy.select(&:green?).select {|b| b.branch_info == 'master' && b.number.to_i >= 4287 }.take(10).sum(&:duration) / 10.0
=> 260.3
```

I hope you all accept this extra time to find easy mistakes such like:

https://github.com/ruby/www.ruby-lang.org/pull/1667/commits/964d1eb71c321c127447bb9d11b69ddde7b0e634#diff-5d067bc5dd97efa0e642d38b7bb88d0f

stomar:
Yes, considering those kind of mistakes happen from time to time, I now tend to think it's a good idea to run those tests. :+1:

PR #1678, opened 2017-10-21
Provide a 'build' task for all languages

wagurano:
`rake build:en` is useful to preview the **English** page.
I want to change the task that builds only one of the languages(**bg de en es fr id it ja ko pl pt ru tr vi zh_cn zh_tw**)
Thank you.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -27,17 +27,28 @@ end
 
 namespace :build do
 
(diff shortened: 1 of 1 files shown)


stomar:
@wagurano Thanks for your contribution!

PR #1682, opened 2017-10-24
Remove Solaris section from installation page (en)

stomar:
[WIP]

The section is very out-of-date.

Unless someone is able to provide current information, I would remove the section completely.

What do you think?


hsbt:
@stomar This section was obsoleted. It should be removed.

PR #1683, opened 2017-10-26
Remove work in progress.

sho-h:
Remove work in progress from jp documentation.


sho-h:
Thanks! @JuanitoFatas @hsbt

PR #1685, opened 2017-11-02
Reorganize release data

stomar:
**[WIP]**

Currently, data on Ruby releases is split into two files: `_data/downloads.yml` provides detailed data for the few Rubies that are currently "featured" on the downloads page, like download URLs and checksums, while `_data/releaes.yml` is a more extensive list of Ruby versions with data for the releases page, like release dates and links to the release announcements.

Duplicating/splitting the data in this way seems clumsy.

Proposed new organization of the data:

* all data is gathered in `releases.yml`, both for the releases and the downloads page
* `downloads.yml` now only defines (by specifying the version number) _which_ Rubies will be listed on the downloads page, the relevant data is taken from `releases.yml`

So, when publishing a new release, the relevant data should be added to `releases.yml`, and the release version added to `downloads.yml`.

One benefit: it seems very unlikely that one of the necessary changes is forgotten, because then the new release would not appear on the downloads page; the releases page would therefore always be up-to-date, other than e.g. reported in #1637 and others.

<del>One downside: the Liquid tags on the downloads page get somewhat complicated, it's not a very elegant solution. (But that might get improved after having a closer look at the Liquid docs.)</del>
_Update: Jekyll provides the `where` Liquid filter, which let's one find items in an array without a loop with condition, so the code is now much clearer._

Acceptable?

@hsbt, @nurse, @unak, @nagachika

**Not ready for merging, because I didn't yet adapt the download pages of the translated sites; they will break.**

### DIFF
--- a/_data/branches.yml
+++ b/_data/branches.yml
@@ -1,7 +1,12 @@
-# name:     branch name (MAJOR.MINOR since 2.1)
-# status:   normal maintenance | security maintenance | eol | preview
-# date:     date of first stable release (YYYY-MM-DD)
-# eol_date: date of EOL (YYYY-MM-DD)
+# This file provides details on the various Ruby branches
+# and their maintenance status.
+#
+# Each entry has the following form:
+#
+# - name:     branch name (MAJOR.MINOR since 2.1)
+#   status:   normal maintenance | security maintenance | eol | preview
+#   date:     date of first stable release (YYYY-MM-DD)
+#   eol_date: date of EOL (YYYY-MM-DD)
 
 - name: 2.4
   status: normal maintenance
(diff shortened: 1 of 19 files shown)


unak:
I'm neutral about this change.
In other word, it's acceptable for me :-)

nagachika:
It seems acceptable for me.

One question. Should I remove the `url` and `sha256` fields in `_data/releases.yml` for the superseded versions when I release new stable version?

stomar:
My thought on this was to always add the data for new releases, and keep the old data untouched. That's simpler, and also it wouldn't hurt to have the data for older releases still around.

PR #1687, opened 2017-11-04
Fix typo in variable assignment demonstration :)

mappuji:
I spot this inconsistency while learning ruby via `Ruby in 20 Minutes` finally I got the source which I suspect it is only a static site and it is open source 😄


stomar:
Thanks for reporting and helping with improving this site. We are always glad about contributions.

Only in this special case I do not see a typo; `a` and `b` are defined so that they can be used in the last calculation, but there is no need in this example to also assign the last result to a variable (it's not reused anywhere).

mappuji:
Thank you, got you

PR #1699, opened 2017-12-18
absent operator → absence operator

znz:
Onigmo use `absence operator` instead of `absent operator` as the name of operator now.


stomar:
Could you provide a reference for this statement?

_PS. Please avoid non-ASCII characters in commit messages._

JuanitoFatas:
> Could you provide a reference for this statement?

I think it's from: https://github.com/k-takata/Onigmo/issues/87

stomar:
Thanks!

PR #1715, opened 2017-12-25
Fix link on 2.5.0 released article

riseshia:
could you review this? @nurse


sorah:
You've found a glitch that made by Dropbox Paper where we used editing collaboration of the release note!

PR #1716, opened 2017-12-26
Use Ruby 2.5.0 and update gem bundle

stomar:
Todo: check that

* generated site does not change in unexpected ways (`diff -r`) _[ok]_
* Travis can handle 2.5 _[ok]_
* Heroku can handle 2.5 _[ok]_


stomar:
Travis:

ˋˋˋ
$ bundle exec rake ci
bundler: failed to load command: rake (/home/travis/build/ruby/www.ruby-lang.org/vendor/bundle/ruby/2.5.0/bin/rake)
LoadError: cannot load such file -- bundler/dep_proxy
ˋˋˋ

hsbt:
It seems travis issues > failed to load command

PR #1773, opened 2018-04-08
Translate CVE-2017-17742: HTTP response splitting in WEBrick news (id)

meisyal:
Another translation that needs to be reviewed by @gozali.


meisyal:
cc: @stomar

PR #1796, opened 2018-05-14
[JA] Fix compare link

riseshia:
Could you check this? @nagachika @unak


nagachika:
Thank you for pointing out this.

PR #1797, opened 2018-05-30
Fix grammar on downloads page (ru)

edenisn:


arbox:
Looks good :+1: Thank you for this correction!

PR #1805, opened 2018-06-13
Fix typo in 2.6.0 released post (zh_cn)

dsh0416:


stomar:
@dsh0416 please use language identifiers („Chinese translation“ is ambiguous).

PR #1817, opened 2018-07-20
Translate 2001 posts (pt)

jcserracampos:
I want to translate all old posts because of history.


stomar:
Commit message is wrong (year) and not in the common format (see new PR title).

I‘m not sure whether adding (very) old news posts makes sense; it doesn‘t hurt of course.

jcserracampos:
@stomar Sorry for the delay.
I updated the commit message.

I want to translate all the old posts for sake of history.
Here, in Brazil, we really love Ruby and when someone will do a talk they can't find historical facts about Ruby in Portuguese.

PR #1819, opened 2018-07-23
Translate 20171225 news (es) and starting glossary

vtamara:


stomar:
Please use separate PR‘s for those two proposals.

Regarding the glossary: I would suggest using the Wiki for that, not the repo itself.

Issue #1827, opened 2018-09-10
Remember language choice


deivid-rodriguez:
Every time I visit ruby-lang.org, I get redirected to https://www.ruby-lang.org/es/, I guess because Spanish is the language of my browser. Since the Spanish translation is always outdated, I have to go to the footer or directly edit the URL to switch to English.

It'd be nice if either my choice would be remembered through a cookie or something, or that the browser language auto-detection was turned off, given that the translations are usually very outdated.

Thanks a lot!

olivierlacan:
The site already has a ton of cookies, I don't see why we couldn't store a language preference as well despite @stomar's objection in https://github.com/ruby/www.ruby-lang.org/pull/1828#issuecomment-427086111. 
![image](https://user-images.githubusercontent.com/65950/48588540-862cee80-e905-11e8-987d-4ab74cd019f0.png)

deivid-rodriguez:
Is there any plan to fix this? I'm happy to try to help myself!

ismailarilik:
I don't agree with the objection. Browsers should open the locale of course. If the language was persistent then changing locale wouldn't mean anything. If you always want the English site then put a bookmark.

deivid-rodriguez:
>  If the language was persistent then changing locale wouldn't mean anything.

What do you mean by this? Of course, the choice should be persistent until you make a different choice in the future. In that case, the next choice should be the new default.

deivid-rodriguez:
This is how multi language sites usually work, since it's user friendly.

PR #1828, opened 2018-09-11
Disable redirect to outdated translations

znz:
I think redirecting to outdated translations is not good.
And I do not want to maintain languages map manually.
So I think it should stop to redirect automatically.
Some languages are commented out in `LANGS_MAP`, because original LANGS does not include them.
Magic number `about_6_months_ago` is to stop redirecting to `pt` which is not included in original LANGS and latest posts are 2017-12-14.

ref  #1827

### DIFF
--- a/_plugins/translation_status.rb
+++ b/_plugins/translation_status.rb
@@ -8,8 +8,27 @@ module Jekyll
   # Outputs HTML.
   module TranslationStatus
 
-    LANGS =  %w{en de es fr id it ja ko ru vi zh_cn zh_tw}
-    START_DATE = '2013-04-01'
+    LANGS_MAP = {
+      "en" => nil,
+      # "bg" => "bg",
+      "de" => "de",
+      "es" => "es",
+      "fr" => "fr",
+      "id" => "id",
+      "it" => "it",
+      "ja" => "ja",
+      "ko" => "ko",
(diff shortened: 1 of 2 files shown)


stomar:
I'm not convinced yet that automatic disabling is a good thing.

Besides that, I think "_plugins/translation_status.rb" is not the right place to define the "languages_json" tag, it should be separate.

deivid-rodriguez:
What about remembering user's language choice on a cookie. Would that be acceptable?

VagishVela:
@deivid-rodriguez I wonder if we could show a banner to ask the user if this the language they want the first time they visit and store it in a cookie? If they change to another language from the footer maybe show them a bar at the top asking if they want to save this as their default language?

If the cookie exists it overrides the auto-redirect.

deivid-rodriguez:
Yeah, that would definitely work for me!

deivid-rodriguez:
Do you have any other proposals to fix #1827, then?

Of course, as a user I can change the language of my browser, or go to the footer and click on English every time I visit the site, or remember to type `ruby-lang.org/en` every time I visit the site. But I think we can agree that it's not a very good user experience. All that assuming that I'm aware that my language translation is outdated. Otherwise I'll just mistakenly assume that the information I'm given is the latest.

PR #1830, opened 2018-09-19
Fix Ruby China links

huacnlee:


JuanitoFatas:
Thanks!

stomar:
@huacnlee Is it by intention `ruby-china.org` but `cache.ruby-china.com`?

huacnlee:
Yes, because .org have some issues, we can’t us it on China CDN, government
internet policy not allow, so, we use .com instead it
Marcus Stollsteimer <notifications@github.com>于2018年9月30日 周日上午2:46写道：

PR #1833, opened 2018-10-02
Translate 2001 posts (pt)

JuanitoFatas:
Revive #1817


jcserracampos:
@JuanitoFatas everything is fine 👍

PR #1836, opened 2018-10-03
Remove MacRuby from about page

sonicdoe:
[MacRuby](https://github.com/MacRuby/MacRuby) hasn’t seen a release since 2012 and has been put “on an indefinite hiatus” in 2015. Therefore, I’d suggest removing it from the about page.

If you’re fine with this change, I can update the translated pages as well.


sonicdoe:
Keep in mind that [RubyMotion](http://www.rubymotion.com) is a commercial product. If the maintainers are okay with featuring such products, I’d gladly open a pull request adding RubyMotion to the about page.

PR #1839, opened 2018-10-04
Fix redirect from http to https on rubybib.org link

VagishVela:
This was redirecting from the http to https version of the website, it is generally best practice to avoid redirects for the user.


stomar:
Thanks for reporting, I'll merge soon.

Issue #1831, opened 2018-09-29
Broken links


prahaladbelavadi:
On the [documentation page](https://www.ruby-lang.org/en/documentation/), the following links are broken or not maintained:
- Ruby Monk
- Hackety

- Ruby Monk shows shows an NGINX 502 Bad gateway error page 
  ![2018-09-29_14-00-17](https://user-images.githubusercontent.com/16673094/46243277-1cf22a00-c3f0-11e8-9de1-deb7f3cff632.png)
- Hackety shows the this error page
  ![2018-09-29_14-02-41](https://user-images.githubusercontent.com/16673094/46243297-60e52f00-c3f0-11e8-82ae-d3d4aef0f7ff.png)

Imo they can be omitted or looked into.

prahaladbelavadi:
This was fixed in an hour :0
Ruby has a new fan now!

PR #1841, opened 2018-10-06
Remove RubyMonk and Hackety Hack links (translations)

meisyal:
As mentioned on #1831, Ruby Monk and Hackety links are broken. I removed that links not only for Bahasa Indonesia but also other languages.
cc: @stomar @hsbt


stomar:
Thanks!

PR #1845, opened 2018-10-14
Fix broken link to MORPHA on Success Stories (en)

VagishVela:
The link was broken and going to a 404 page, so changing to point to the
root of the domain.

[Success Stories](https://www.ruby-lang.org/en/documentation/success-stories/)

### DIFF
--- a/bg/documentation/success-stories/index.md
+++ b/bg/documentation/success-stories/index.md
@@ -62,7 +62,7 @@ lang: bg
 [2]: http://www.motorola.com
 [3]: http://www.sketchup.com/
 [4]: http://www.torontorehab.com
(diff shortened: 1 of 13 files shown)


meisyal:
Hi @VagishVela,

Maybe I am wrong. I think the link for MORPHA project is not https://www.morpha.de because MORPHA project is a reactive control part for the Siemens service robot, as described on success stories. That website doesn't describe that.

It should be somewhere else. I can't find one. This [research paper] maybe helps you.

[research paper]: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.19.8130&rep=rep1&type=pdf

JuanitoFatas:
Need a rebase 😉

PR #1849, opened 2018-10-15
Simplify the preview local instructions

VagishVela:
We can use jekyll's in built features to display the website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,19 +53,13 @@ your commits and PRs should look like.
 
 ### Preview Locally
 
-Generate the website with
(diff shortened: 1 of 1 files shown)


stomar:
Two reasons against this proposal:

* when served with `jekyll` the site does not behave in the same way as in production (with `rackup`), where additional rack middleware is present (e.g. for redirections)
* `jekyll serve` always rebuilds the site, which takes a _very_ long time, so its very useful to be able to preview with an existing build

And having both options in the instruction would be no improvement regarding simplicity.

PR #1852, opened 2018-10-16
Update link to Toronto Rehab (translations)

meisyal:
I updated Toronto Rehab's link from http://www.torontorehab.com to https://www.uhn.ca/TorontoRehab because the previous one is no longer available. cc: @stomar @hsbt


JuanitoFatas:
Thanks!

PR #1864, opened 2018-10-20
Update RDoc URL for RDoc itself

tadd:
I found that RDoc URL of RDoc itself is too old in many pages.
http://docs.seattlerb.org/rdoc/ uses RDoc 4.1.1, but latest version https://rubygems.org/gems/rdoc is 6.0.4.

I heard current canonical URL https://ruby.github.io/rdoc from @aycabta, a maintainer of RDoc, so I substituted with it using `sed -i`.


aycabta:
:+1:

PR #1875, opened 2018-11-08
%s/메소드/메서드/g

riseshia:
Ref: http://www.korean.go.kr/front/onlineQna/onlineQnaView.do?mn_id=61&qna_seq=11976

From https://github.com/ruby/www.ruby-lang.org/pull/1872#discussion_r231783436

Could you check this? @ruby/www-ruby-lang-org-i18n-ko


stomar:
Please use descriptive commit messages in English and ASCII, not all maintainers understand Korean, and not all shells can display arbitrary characters.

riseshia:
Oops, I'm sorry to this. do you want to revert and re-merge with new commit message?

Issue #1606, opened 2017-07-04
Spanish translations


gargola:
Hello @stomar ,

I've noticed that the spanish translations are very out-dated and would like to collaborate with the project by fixing that. I can start right away so, please let me know what the best approach for this would be.

-gargola

dabit:
Hi @gargola, I personally apologize for this since I've been up to my neck in work to actually make them. I'm in charge of the spanish translations.

If you send the PR, I'll review and merge it.

Thanks in advance!

gargola:
No worries @dabit, through collaboration we can make it happen. That's the beauty of Open Source 😄  I'll start working on this right away so expect a PR pretty soon.

Thanks a lot for the prompt reply!

Regards.

Quintus:
Now that I look into it. Cf. #1893. Looks as if there's hope for the `es` team. /cc @hsbt again

PR #1876, opened 2018-11-12
Translation of 2018-06-20-support-of-ruby-2-2-has-ended (es)

vtamara:
Helping #1606


vtamara:
@gfvcastro, @dabit, @vic, @soveran could you please review?

PR #1888, opened 2018-11-21
The announcement for 

hsbt:

PR #1899, opened 2018-11-30
Fix code snippets in snap post

havenwood:
This fixes the code snippets in #1888, since double-dashes were not appearing correctly.

Before:
```
sudo snap install ruby –classic –channel=2.4/stable
```

After:
```
sudo snap install ruby -–classic -–channel=2.4/stable
```


JuanitoFatas:
Awesome! Great fix. Thanks Shannon!

hsbt:
@havenwood Thanks!

PR #1900, opened 2018-11-30
Replace "..." with "and so on" in FAQ 1

nsuchy:
This pull request replaces the text "..." with "and so on" so the two sentences about OO features are consistent with one another. Thoughts?


JuanitoFatas:
I think this is great, thanks Nathaniel!

PR #1904, opened 2018-12-03
Translate snap post (zh_cn)

dsh0416:


olivierlacan:
Same as https://github.com/ruby/www.ruby-lang.org/pull/1906#issuecomment-443984451 it would be a good idea to add Markdown links around the URLs in this post.

dsh0416:
@olivierlacan Thank you for your review. I have followed up my commit with #1907

JuanitoFatas:
Awesome!

PR #1933, opened 2018-12-26
Make font size of code the same as of parent element

ioquatix:
Let code be same size as parent containing element. Someone should check this before deploying to production.

### DIFF
--- a/stylesheets/main.css
+++ b/stylesheets/main.css
@@ -135,8 +135,6 @@ code {
   border: 1px solid #ddd;
   background-color: #f8f8f8;
   font-family: Consolas, "Liberation Mono", Courier, monospace;
-  font-size: 14px;
-  line-height: 16px;
   text-shadow: 0 1px 0 #fff;
   -webkit-border-radius: 3px;
      -moz-border-radius: 3px;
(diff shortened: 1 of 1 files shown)


stomar:
I cannot preview right now, but in principle +1.

stomar:
I applied the first commit (changing the font size for code) in e72d59b11ef642478c3ceacc2b66490ee20262e5.

I previewed some posts and pages and it seems to be fine; the code samples on the root page are not affected; and it‘s a big improvement for subheadings including code snippets. 👏

I‘m not sure about the second commit (adjusting the line heights). That is, _how_ it is implemented:

* it would affect „normal“ paragraphs and the page footer, but not the summaries for the download, community, ... pages or the definition lists that are used in many of these pages
* I think it would be better to keep definitions of line-height together with definitions of font-size, or a size changed in a comletely different part of the CSS might lead to unexpected results
* I‘m not sure about using a relative size together with font-sizes that are all specified using absolute sizes

But, in principle, I think increasing the line-height a bit would not hurt. I suggest opening a different PR or issue.

stomar:
It‘s not deployed yet, for technical reasons I cannot deploy currently.

PR #1934, opened 2018-12-26
Remove duplicate `status` and use expected `date` key for 2.6 branch

epistrephein:
Ruby 2.6 release date is listed with the `status` key instead of the `date` key in the branches page, with the result that the branch actual status is not shown and the release date is in the wrong position.

This PR fixes the issue, restoring the correct keys.


nurse:
thanks!

PR #1928, opened 2018-12-25
 Translate Ruby 2.6.0 Release (ko)

riseshia:
Ref: https://github.com/ruby/www.ruby-lang.org/issues/1894

This translates https://github.com/ruby/www.ruby-lang.org/pull/1926

Check 0dd4036 for review, @ruby/www-ruby-lang-org-i18n-ko :pray:

PR #1935, opened 2018-12-26
Follow up 43399ee (ko)

riseshia:
From: https://github.com/ruby/www.ruby-lang.org/pull/1928#discussion_r243998800

FYI @ruby/www-ruby-lang-org-i18n-ko


yous:
Thanks!

PR #1936, opened 2018-12-26
Update about/website/index.md with supporting orgs (pt)

vnbrs:
This PR updates the list of supporting organizations for the Portuguese i18n file, based on the English one


stomar:
For easier side-by-side diff with `en`, could you list references in the same order (rubyassociation)?

vnbrs:
> For easier side-by-side diff with `en`, could you list references in the same order (rubyassociation)?

I've just changed it to match the en version. =D

stomar:
Thanks!

PR #1958, opened 2019-02-02
Remove defunct conference sites (en)

jacobherrington:
Some of these conference sites have been repurposed or are no longer
hosted. I removed those, but left some conferences that are no longer
active, as long as their sites were still accurate.

If this change is merged, I will add some newer Ruby conferences to list.
e.g. [Southeast Ruby](https://southeastruby.com/)


stomar:
Applied as da7c540, without updating reference numbers to avoid „git churn“.

PR #1965, opened 2019-02-08
Editor list fixes (en)

jacobherrington:
I will make these changes on other translations if this is considered a good change.


JuanitoFatas:
This is great, thank you! 💯

PR #1966, opened 2019-02-08
Update dmoz.org to dmoztools.net (en)

jacobherrington:
dmoz.org is no longer operating, but the information is still available
on dmoztools.net. I think it's worth considering changing these
resources out for more modern ones.

Again, this is a change I can make across more translations if it's
considered valuable.

### DIFF
--- a/en/community/index.md
+++ b/en/community/index.md
@@ -50,5 +50,5 @@ General Ruby Information
 
 
 [3]: http://rubycentral.org/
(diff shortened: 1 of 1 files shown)


JuanitoFatas:
Thanks ❤️ 

> Again, this is a change I can make across more translations if it's
considered valuable.

Feel free to submit new Pull Requests 🎁

PR #1969, opened 2019-02-09
Fix dmoz.org links

jacobherrington:
dmoz.org is no longer active. dmoztools.net still hosts a lot of the
same content. This commit updates those links.

This is a continuation of #1966 per @JuanitoFatas.


JuanitoFatas:
This is great, thank you!

PR #1980, opened 2019-02-23
Translate missing parts in code examples (tr)

ismailarilik:
Here specifies that code examples shouldn't be translated  with an exception of comment lines: https://github.com/ruby/www.ruby-lang.org/wiki#notes-for-translators
But just translating comments, explaining code to a native and newby person is not possible; translating strings, variables and other names is also necessary.

I am not opening an issue for this because as I have seen, some changes had already been made to these examples, which violates this rule. But if it is necessary, I can open an issue and we can continue discussing the issue there. If my proposal OK, then the line in the wiki I mentioned should be edited (or deleted). I can do that, too.

### DIFF
--- a/tr/examples/cities.md
+++ b/tr/examples/cities.md
@@ -4,19 +4,19 @@ layout: null
 
 {% highlight ruby %}
 # Tüm bir Array üzerinde
-# matematik işlem yapmak
+# matematik işlemi yapmak
 # isteseniz dahi
 # Ruby ne istediğinizi
 # hemen anlar
-cities  = %w[ London
-              Oslo
-              Paris
-              Amsterdam
-              Berlin ]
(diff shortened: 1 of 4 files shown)


tr/examples/cities.md
```diff
@@ -4,19 +4,19 @@ layout: null
...
-              Paris
-              Amsterdam
-              Berlin ]
-visited = %w[Berlin Oslo]
+cities = %w[ London
+             Oslo
+             Paris
+             Amsterdam
+             Berlin ]
+visited  = %w[Berlin Oslo]
```
enderahmetyurt: fazla space var

ismailarilik: Good catch! I have fixed it.

PR #1981, opened 2019-02-23
Change string in code to reflect correct output (en)

ismailarilik:


ismailarilik:
The same fix for translations is necessary.

hlcfan:
This wouldn't be an issue 😺. As it calls `#capitalize` on `name` while initializing the object.
https://repl.it/repls/LividCautiousConditional

ismailarilik:
😄 You are right, I didn't notice.

PR #1987, opened 2019-03-03
Fix markup issues

spk:
Hi just fixing some markup errors found while working on https://github.com/ruby/www.ruby-lang.org/pull/1420 in case this is not merged; cheers

### DIFF
--- a/admin/translation-status/index.html
+++ b/admin/translation-status/index.html
@@ -1,6 +1,3 @@
----
----
-
 <!DOCTYPE html>
 <html lang="en">
 <head>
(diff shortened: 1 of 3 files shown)


admin/translation-status/index.html
```diff
@@ -1,6 +1,3 @@
----
----
-
```
znz: Why do you remove [YAML front matter](https://jekyllrb.com/docs/front-matter/)?

If you show the file directly, I recommend to use rake commands described in [Preview Your Changes](https://github.com/ruby/www.ruby-lang.org#preview-your-changes).

spk: Fixed in https://github.com/ruby/www.ruby-lang.org/commit/d0cbdb32665864929c618ac1c80e223af31250c3 thanks

stomar:
@spk Thanks for catching those errrors! But please avoid silently introducing unrelated changes.

Also thanks to @znz.

spk:
Sorry again I did not know fixed in https://github.com/ruby/www.ruby-lang.org/pull/1988 by @znz thanks

PR #1990, opened 2019-03-04
Validate with libtidy on travis

spk:
ref https://github.com/ruby/www.ruby-lang.org/pull/1420

Use libtidy https://github.com/htacg/tidy-html5 to check the build-ed `_site` generated on CI.
Also update validate-website gem.

Time ~ 10sec for `check:markup` cf https://travis-ci.org/ruby/www.ruby-lang.org/builds/504098647#L590

### Why?

Since #832 we can see that errors happen on markup even it we are using markdown on most of the pages (cf #1418, #1458, #1987 )

The main blocker before was that it increase build time but the complete website check take now ~10sec to ~12sec

cc @stomar @hsbt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,9 @@ rvm: 2.6.2
 cache: bundler
 env:
   - TASK=ci
-before_install: gem install bundler:2.0.1
+before_install:
+  - gem install bundler:2.0.1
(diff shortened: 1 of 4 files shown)


JuanitoFatas:
We are not using Travis anymore, thank you!

PR #1993, opened 2019-03-06
Paralleling Travis jobs to shorten build duration

alaxalves:
Fixed the in-between travis build caching problem and makes the jobs parallel making easier to identify an eventual issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,11 @@ sudo: false
 rvm: 2.6.1
 cache:
   bundler: true
-script: bundle exec rake ci
+env:
+  matrix:
+    - SCRIPT="bundle exec rake build"
+    - SCRIPT="bundle exec rake test"
+script: $SCRIPT
 # Notifications, used by our Gitter channel.
 notifications:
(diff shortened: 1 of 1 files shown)


stomar:
Could you please shortly explain the „in-between travis build caching problem“?

alaxalves:
> Could you please shortly explain the „in-between travis build caching problem“?

When you add caching like this:
``` yml
cache:
  bundler: true
```
to Travis you must specify a directory to cache as specified [here](https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies). This was causing Travis not cache properly sometimes. Since ruby-lang does not specify a different folder to install dependencies just using `cache: bundler` to the yml fixes it.

stomar:
I added a commit with a simpler config: `matrix` is not necessary here (no global env vars), and using TASK instead of SCRIPT makes the log more readable, IMO.

What do you think?

alaxalves:
@stomar I think it's great. Thank you for the feedback.

PR #1997, opened 2019-03-10
Add a link to rurema for Ruby 2.6.0 (ja)

sho-h:
update rurema link to add 2.6.0 (ja).


znz:
Now, older versions are upper, but how about reversing links?

sho-h:
Thanks! I changed order.

PR #2006, opened 2019-03-15
Small improvements in 2.5.5 post (en)

stomar:


stomar:
@nagachika

nagachika:
@stomar Thank you for your correction! 
I have one question. I wanted to say `the application using fork from multi-threaded parent process` in `multi-thread+multi-process application`. Does the notation `multi-thread/multi-process` have the same nuance?

stomar:
@nagachika Sorry for the late response. Actually both „multi-thread+multi-process“ and „multi-thread/multi-process“ are very unspecific, so I added your more detailed explanation in a4a0ad3589d601154fd70ac86abeaa64459b31d5.

PR #2020, opened 2019-03-27
Make other implementations more visible to users

brunoborges:
The list of other implementations is hidden in the About page, which few developers will hit. By showing developers this list in the Downloads page, developers can be better informed and aware that the Ruby language is rich and widely adopted to the point of having multiple implementations and engines to run Ruby code.

### DIFF
--- a/en/downloads/index.md
+++ b/en/downloads/index.md
@@ -68,6 +68,40 @@ or previews, see the [Releases page][releases].
 Information about the current maintenance status of the various
 Ruby branches can be found on the [Branches page][branches].
 
+### Other Implementations of Ruby
+
+Ruby, as a language, has a few different implementations.
+This page has been discussing the reference implementation, in the
+community often referred to as **MRI** (“Matz’s Ruby Interpreter”)
+or **CRuby** (since it is written in C), but there are also others.
+They are often useful in certain situations, provide extra
+integration to other languages or environments, or have special features
+that MRI doesn’t.
+
+Here’s a list:
+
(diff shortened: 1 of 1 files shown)


brunoborges:
Hey @ismailarilik, @hsbt, @stomar -- any chance you can take a look at this?

sorah (changes requested):
The download page is for MRI. I guess there's better room for this kind of information. Although I agree to have a link to the information on download page. 

Current changes are too big for putting onto the download page.

sorah:
OK, this language was exactly taken from the existing about page. https://www.ruby-lang.org/en/about/

The download page isn't for this language.

brunoborges:
Let's forget about the language for a second, and discuss the main point: Ruby is a programming language specification that has multiple implementations. The proposal here is to surface these other implementations under the Download page because that is the page that almost every developer visits.

I'd be curious to see a traffic analysis of the Download page versus the About page, but I'd confidently guess that the About page gets less than a 1/10 of the hits that the Download page gets.

By raising awareness that Ruby has multiple implementations, developers can more confidently invest in Ruby for new projects. This of course assuming that ruby-lang.org is a website dedicated to the Ruby language more than to the Ruby runtime.

sorah:
But ruby-lang.org is for MRI. Why you put the info above of Compiling Ruby section?

brunoborges:
Got it, thanks for the feedback. I'll work on that proposed change sometime this week.

PR #2026, opened 2019-04-04
Fix the download section of 2.4.6 release (en)

JuanitoFatas:
Current

Screenshot 2019-04-04 20 03 51

After

Screenshot 2019-04-04 20 03 40


JuanitoFatas:
I take the liberty of merging this since it's broken right now.

Issue #1973, opened 2019-02-18
Broken links


dotemacs:
On this page:

https://www.ruby-lang.org/en/documentation/ruby-from-other-languages/to-ruby-from-java/

The links to WxRuby & FXRuby are broken.

stomar:
fxRuby still gets updates, see https://github.com/larskanis/fxruby/tree/1.6, maybe we should link to the repo or inquire there about the website.

PR #2035, opened 2019-04-11
Use github page for wxruby fxruby

spk:
Fix https://github.com/ruby/www.ruby-lang.org/issues/1973


JuanitoFatas:
This is great, thank you! @spk <3

Quintus:
OK for de/

PR #2046, opened 2019-04-21
Russian version of rubyinstaller.org website link

aristofun:
It's reasonable and natural to see links to localized russian version of RubyInstaller website on russian pages of ruby-lang.org

### DIFF
--- a/ru/downloads/index.md
+++ b/ru/downloads/index.md
@@ -73,4 +73,4 @@ lang: ru
 [mirrors]: /en/downloads/mirrors/
 [rvm]: http://rvm.io/
 [rbenv]: https://github.com/rbenv/rbenv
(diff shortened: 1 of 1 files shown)


stomar:
The .ru site doesn‘t seem to be related in any official way to rubyinstaller.org, so we should _not_ link to it.

aristofun:
It is "officially" and publicly forked & available on github (what is officially encouraged by .org authors)
and seems like professionally translated: https://github.com/installero/rubyinstaller.ru

Please reconsider to encourage this contribution of .ru authors

stomar:
That doesn‘t mean anything, anybody can create his own fork.

We at ruby-lang.org can not guarantee that the provided binaries are unmodified and will be so also in the future, so we should still link to the original site.

PR #2047, opened 2019-04-21
Russian version of rubyinstaller website

aristofun:
Please update the link to match the language of `ru/installation/` page as well


stomar:
The .ru site doesn‘t seem to be related in any official way to rubyinstaller.org, so we should _not_ link to it.

aristofun:
What would make it more (or less) "official" than rubyinstaller.org?

aristofun:
It is "officially" and publicly forked & available on github (what is officially encouraged by .org authors)
and seems like professionally translated: https://github.com/installero/rubyinstaller.ru

Please reconsider

stomar:
That doesn‘t mean anything, anybody can create his own fork.

We at ruby-lang.org can not guarantee that the provided binaries are unmodified and will be so also in the future, so we should still link to the original site.

PR #2061, opened 2019-05-08
Follow update translation (ko)

riseshia:
d727a63 for a4a0ad3
07f3533 for #2056

Could you review this? @ruby/www-ruby-lang-org-i18n-ko


riseshia:
:pray:

PR #2065, opened 2019-05-16
Remove reference to rubyforge.org in ja libraries page

okuramasafumi:
`rubyforge.org` has been inaccessible for years and I don't think it's worth mentioning here.
Also, GitHub is now quite popular so I changes the wordings slightly.


okuramasafumi:
Pushed!

PR #1680, opened 2017-10-23
Drop mention of RubyForge from libraries page (en)

stomar:
Fixes #1668.

@MischaTheEvil good enough for you?

PR #2068, opened 2019-05-18
Remove reference of RubyForge from libraries page (ko)

yous:
Ref #2065, #1680.


JuanitoFatas:
Thanks! 👍

yous:
Thanks!

PR #2110, opened 2019-07-17
Add regional USA conference for ruby (en)

VagishVela:
This is a USA event, so only added it to the english page for conferences.


JuanitoFatas:
Thanks!

PR #2129, opened 2019-08-29
Use Jekyll 4

hsbt:
Jekyll 4 was improved the generation time for markdown.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,11 +2,11 @@ source "https://rubygems.org"
 ruby ENV['CUSTOM_RUBY_VERSION'] || '~> 2.6.3'
 
 gem 'rake'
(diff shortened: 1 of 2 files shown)


mame:
@znz said that updating rouge requires updating css.  Need to confirm if it is rendered correctly.

jcserracampos:
@mame I was thinking... Can we make a preview pipeline to new PRs?
Something like netlify does.

hsbt:
We can use heroku preview app for this.

jcserracampos:
@hsbt It'd incredible because we can catch more errors when seeing live pages.

jcserracampos:
I found no problems with render.
LGTM.

hsbt:
@jcserracampos Thanks!

PR #2141, opened 2019-09-10
Fixing typo in PT documentation

douglasgabriel:


lex111:
@douglasgabriel obrigado! :smile:

PR #2149, opened 2019-10-02
Nightly snapshot is from git instead of svn now

znz:
- stable-snapshot too, but it may be from svn again, so should not mention
- commit email to ruby-cvs too


JuanitoFatas:
Thanks @znz and everyone who reviewed!! <3

Issue #2143, opened 2019-09-19
Link to 'Ruby C API Reference' is broken


mike-shock:
On the page https://www.ruby-lang.org/en/documentation/ the hyperlink 'Ruby C API Reference' leads to the page (https://docs.ruby-lang.org/en/trunk/extension_rdoc.html) which IS NOT FOUND!

PR #2158, opened 2019-10-07
Fix doc link to master instead of trunk

hahmed:
Fixes https://github.com/ruby/www.ruby-lang.org/issues/2143


JuanitoFatas:
Thanks @hahmed 🙇

PR #2159, opened 2019-10-07
Fix Travis CI build

yous:
Maybe https://github.com/ruby/www.ruby-lang.org/commit/8bdc957da6502624181fefff72e3a041ff8c69cf caused the failure.


JuanitoFatas:
👍

PR #2161, opened 2019-10-08
Removing ODS from success stories

esparta:
According to [archive.org][1] ODS is not longer on business and about Feb 2019 their website started to return 403

[1]: https://web.archive.org/web/20190226032335/http://www.ods.org/


JuanitoFatas:
:100:

esparta:
Thanks!

PR #2171, opened 2019-10-16
Remove the links of Lucent (no longer existing website)

VagishVela:
Lucent's website no longer exists, and has since merged with Alcatel.


yous:
Maybe the whole section should be removed.

PR #2179, opened 2019-10-20
Update the Silicon Valley Ruby Conference section on Conferences

VagishVela:
SVForum doesn't exist anymore, and the conference ran for two years
only. So updating the context, might be useful for anyone who might want
to restart the conference.

### DIFF
--- a/en/community/conferences/index.md
+++ b/en/community/conferences/index.md
@@ -43,8 +43,9 @@ event dates, location, CFP (Call For Proposals) and Registration information.
 to offset expenses for local and regional groups wanting to organize
 events.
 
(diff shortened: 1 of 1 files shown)


meisyal:
Hi @VagishVela, could you also update the other language?

yous:
The build is failing because of trailing whitespace.

stomar:
I resolved the conflict and committed 9f4e6bc2abf8f2678ad87562aebd21e7ff3ccfbb and 91e5be4a355511bb3522f96ce8be353439264ed8.

Regarding the translations: since there is not only the removed link but also a change of the content, I would leave that to the translation teams.

meisyal:
OK @stomar

PR #2197, opened 2019-10-22
Fix trailing space in 2.7.0-preview2 post (ja)

JuanitoFatas:
This caused the build to fail.


JuanitoFatas:
I will fix the commit message, thanks @jcserracampos!

PR #2250, opened 2019-11-03
Lock Jekyll to 4.x and switch back to upstream lanyon

stomar:
* Use the upstream `lanyon` gem again, which was updated to drop the constraint on Jekyll versions and now also allows Jekyll 4 (or higher); the new version doesn't include other noteworthy changes.
* Lock Jekyll to version 4.x, updates should only be done intentionally and after thorough testing.


stomar:
@hsbt Sorry for the delay, I had to take a "sabbatical" from open source work, so to speak; I hope I can contribute a bit more again in the next time.

PR #2251, opened 2019-11-04
Send FAQ as is (tr)

ismailarilik:
Send FAQ as is because I don't have time to translate it. I may translate it in the future or someone else may do. Until then redirecting tr site to en site is not necessary and means waste of work because all pages and news are translated except FAQ pages.


stomar:
Please link to the `en` version directly, instead of duplicating the content. That's much less maintenance effort.

PR #2252, opened 2019-11-04
Update config file for translated content (tr)

ismailarilik:


stomar:
Conflict due to #2261 which fixed accidentally introduced wrong line breaks (CR/LF).

stomar:
Sorry I have to close, but the PR cannot be fixed in this way. Merge commits in feature branches are an absolute "no-no" (they extremely clobber the commit history), which is also mentioned in the Wiki. The correct way would have been to rebase the feature branch onto master and either force push the rebased branch, or open a new PR with it.

PR #2260, opened 2019-11-05
Specify that tr language is maintained

ismailarilik:
All content except FAQ is up-to-date now.

FAQ may be translated in the future. Until then, it is not logical not to show tr content to tr people.

### DIFF
--- a/index.html
+++ b/index.html
@@ -11,6 +11,7 @@
         "ja":    "ja",
         "ko":    "ko",
         "pt":    "pt",
+        "tr":    "tr",
         "zh-CN": "zh_cn",
         "zh-TW": "zh_tw"
       };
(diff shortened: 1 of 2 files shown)


stomar:
The "unmaintained" status requires a team of reviewers, since according to experience otherwise it's very hard to keep up with news post translations for a longer period of time. Are there others?

ismailarilik:
> The "unmaintained" status requires a team of reviewers, since according to experience otherwise it's very hard to keep up with news post translations for a longer period of time. Are there others?

Of course; we are a team.

stomar:
> Of course; we are a team.

I had only seen your contributions, so...

ismailarilik:
> > Of course; we are a team.
> 
> I had only seen your contributions, so...

Yes, only I contributed because @enderahmetyurt is busier than me. But he reviewed almost all of my contributions. So we had a team work.

stomar:
Thanks for the effort!

PR #2270, opened 2019-11-08
Translate 2019-10-01 news (ko)

yous:
Ref: https://github.com/ruby/www.ruby-lang.org/compare/1e430491bd...72097faf35, #2147, #2156

The actual diff: https://github.com/ruby/www.ruby-lang.org/commit/67326d2c682efdb405cb63af5de08df5a76cc470


yous:
Thank you!

PR #2272, opened 2019-11-08
Add GitHub actions workflow for CI tests

stomar:
Use GitHub actions workflow for running `rake ci`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,13 +4,15 @@ on: [push, pull_request]
 
 jobs:
   build:
-    name: build (${{ matrix.ruby }})
+    name: build (${{ matrix.ruby }} / ${{ matrix.os }})
 
     strategy:
       matrix:
         ruby: [2.6.3]
+        os: [ubuntu-latest, macos-latest]
+      fail-fast: false
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
     env:
       # override the Ruby version specified in the Gemfile
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,36 @@
+name: ci
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: build (${{ matrix.ruby }})
+
+    strategy:
+      matrix:
+        ruby: [2.6.3]
+
+    runs-on: ubuntu-latest
+
+    env:
+      # override the Ruby version specified in the Gemfile
+      CUSTOM_RUBY_VERSION: ${{ matrix.ruby }}
+
+    steps:
+      - name: Dump environment
+        run: env | sort
+      - name: Checkout ruby/www.ruby-lang.org
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      - name: Setup Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          architecture: x64
+      - name: Install bundler
+        run: gem install bundler --no-document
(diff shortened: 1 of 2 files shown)


stomar:
**Note: WIP**

My first trial with GitHub actions.

It seems to work, but compared to Travis takes forever (installation of the gem bundle takes about 4 to 5 minutes),

stomar:
I added caching for the gem bundle, now took 1 min 20 s. (BTW, Travis after 5 min still pending...)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+      - '*'
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        ruby: [ '2.6.3' ]
```
JuanitoFatas: 2.6.5?

https://github.com/ruby/www.ruby-lang.org/blob/cfcb63e3264df8dba8dec7138b09e663ef540f40/.travis.yml#L3

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          architecture: 'x64'
+      - name: Cache gem bundle
+        uses: actions/cache@v1
+        with:
+          path: vendor/bundle
+          key: bundle-${{ hashFiles('**/Gemfile.lock') }}
+      - name: Install bundler
+        run: gem install bundler
```
JuanitoFatas: would `--no-document` help here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        ruby: [ '2.6.3' ]
+      fail-fast: true
+    runs-on: ubuntu-latest
+    env:
+      CUSTOM_RUBY_VERSION: ${{ matrix.ruby }}
```
JuanitoFatas: What's this for? 🙋‍♂

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+    name: Ruby ${{ matrix.ruby }}
+    steps:
+      - name: Dump environment
+        run: env | sort
+      - name: Checkout ruby/www.ruby-lang.org
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
```
JuanitoFatas: Awesome 🥇

.github/workflows/ci.yml
stomar: See above; 2.6.5 currently is not provided by the setup-ruby action.

JuanitoFatas: 🤦‍♂  I see. Sorry.

.github/workflows/ci.yml
JuanitoFatas: To work around GitHub Actions not supported 2.6.5.

.github/workflows/ci.yml
stomar: Not much :smile:, but a good suggestion nevertheless.

.github/workflows/ci.yml
stomar: To overide the default Ruby version in the Gemfile (2.6.5); else it would abort.

stomar:
I do not know yet why I have to use `hashFiles('**/Gemfile.lock')` instead of simply `hashFiles('Gemfile.lock')` (which does not work). I'm still investigating.

stomar:
I'm confident enough about this to merge and test under real conditions.

hsbt:
@stomar Thanks.

PR #2273, opened 2019-11-10
Update Silicon Valley Ruby Conference section (tr)

ismailarilik:
9f4e6bc2abf8f2678ad87562aebd21e7ff3ccfbb
91e5be4a355511bb3522f96ce8be353439264ed8


stomar:
The commit message makes no sense (this commit message is taken from the second commit, that only makes a small fix to the previous one).

PR #2293, opened 2019-12-03
Add a note: Separation of positional and keyword arguments in Ruby 3.0

mame:


en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+bar("key" => 42, :sym => 43)
+  #=> Ruby 2.6: [{"key"=>42}, {:sym=>43}]
+  #=> Ruby 2.7: [1, {"key"=>42, :sym=>43}]
+
+# Use braces to keep the behavior
+bar({"key" => 42}, :sym => 43)
+  #=> Ruby 2.6 and 2.7: [{"key"=>42}, {:sym=>43}]
+```
+
+Ruby 2.7 will still split hashes with a warning if passing a Hash or keyword arguments with both Symbol and non-Symbol keys to a method that accepts explicit keywords but no keyword rest argument (`**kwargs`).  This splitting will be removed in Ruby 3, and an `ArgumentError` will be raised.
```
sorah:
```suggestion
Ruby 2.7 still splits hashes with a warning if passing a Hash or keyword arguments with both Symbol and non-Symbol keys to a method that accepts explicit keywords but no keyword rest argument (`**kwargs`).  This splitting will be removed in Ruby 3, and an `ArgumentError` will be raised.
```

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+foo(**empty_hash)
+  #=> Ruby 2.6 or before: {}
+  #=> Ruby 2.7: warning: The keyword argument is passed as the last hash parameter
+  #             warning: for `foo' defined here
+  #=> Ruby 3.0: ArgumentError: wrong number of arguments
+```
+
+### 3. The no-keyword-arguments syntax (`**nil`) is introduced
+
+`**nil` is allowed in method definitions to explicitly mark that the method accepts no keyword arguments. Calling such a method with keyword arguments will result in an `ArgumentError`.  (This is not an incompatibility but a new feature.)
```
sorah:
```suggestion
You can use `**nil` in a method definition to explicitly mark the method accepts no keyword arguments. Calling such methods with keyword arguments will result in an `ArgumentError`. (This is actually a new feature, not an incompatibility)
```

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+
+```
+def foo(*args, **nil)
+end
+
+foo(k: 1)
+  #=> Ruby 2.7 or later: no keywords accepted (ArgumentError)
+```
+
+This is useful to make it explicit that the method does not accept keyword arguments.  Otherwise, the keywords are absorbed in the rest argument in the above example.  If you want to extend the method to accept a keyword argument in future, it may break the compatibility.
```
sorah:
```suggestion
This is useful to make it explicit that the method does not accept keyword arguments.  Otherwise, the keywords are absorbed in the rest argument in the above example.  If you extend a method to accept keyword arguments, the method may have incompatibility as follows:
```

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+# If you want to keep the behavior in Ruby 3.0, write braces to make it an
+# explicit Hash
+bar({ k: 42 }) # => {:k=>42}
+```
+
+## What was deprecated?
+
+In Ruby 2, keyword arguments can be treated as the last positional Hash argument and a last positional Hash argument can be treated as keyword arguments.
+
+Because the automatic conversion was sometimes too complex and troublesome (described in the last section), it is deprecated in Ruby 2.7 and will be removed in Ruby 3.  In other words, keyword arguments will be completely separated from positional ones.  If you want to pass keyword arguments, you should always use `foo(k: expr)` or `foo(**expr)`.  If you want to accept keyword arguments, in principle you should always use `def foo(k: value)` or `def foo(k:)` or `def foo(**kw)`.
```
sorah: What is "(described in the last section)?" I don't think the last section "Typical cases" explaining the ambiguous cases clearly.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+# If the method is extended to accept a keyword
+def foo(*args, mode: false)
+  p args
+end
+
+# The existing call may break
+foo(k: 1) #=> ArgumentError: unknown keyword k
+```
+
+## Why was it deprecated?
```
sorah:
```suggestion
## Why we're deprecating the automatic conversion
```

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+foo(h)
+  # => demo.rb:11: warning: The last argument is used as the keyword parameter
+  #    demo.rb:2: warning: for `foo' defined here; maybe ** should be added to the call?
+  #    42
+
+# If you want to keep the behavior in Ruby 3.0, use double splat
+foo(**h) #=> 42
+```
+
+Here is another case.  You can use braces to pass a Hash instead of keywords explicitly.
```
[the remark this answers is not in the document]
mame: I hear that "another" includes "an".  Actually, google says:

* "Here is another case": 2M hits
* "Here is an another case": 7k hits

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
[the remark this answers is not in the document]
mame: Could you remove the last empty backquotes?

mame: I meant the final section "Why we're deprecating the automatic conversion", not the previous one.  Is "described in the final section" better?

mame:
@sorah Thank you very much for your review!

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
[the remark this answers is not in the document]
jeremyevans: `another case` is definitely better than `an another case`.  You could also use `an additional case`, but that's too wordy IMO.  Showing the braces seems fine, though.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
jeremyevans: Yes, you definitely want to use `final section` instead of `last section`, as `last section` implies the section you just read, not the section at the end of the document.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+
+## Will my code break on Ruby 2.7?
+
+A short answer is "maybe not".
+
+The changes in Ruby 2.7 are designed as a migration path towards 3.0.  In principle, Ruby 2.7 only warn against the behavior that will not work on Ruby 3.0.  But there are some incompatible changes in Ruby 2.7, which we consider minor.  See the "Other minor changes" section for details.
+
+Except for the warnings and minor changes, Ruby 2.7 attempts to keep the compatibility with Ruby 2.6.  So, your code will probably work on Ruby 2.7, though it may emit warnings.  And by running it on Ruby 2.7, you can check if your code is ready for Ruby 3.0.
+
+The warnings are printed every time the deprecated behavior is triggered.  You may think it is annoying.  You can suppress duplicate warnings using the following hack:
```
jeremyevans: This section needs to be updated now that duplicate warnings are suppressed.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+
+In short: use `Module#ruby2_keywords` again.
+
+```
+ruby2_keywords def foo(*args, &block)
+  target(*args, &block)
+end
+```
+
+Unfortunately, we need to use the old-style delegation (i.e., no `**kwargs`) because Ruby 2.6 or prior does not handle the new delegation style correctly.  This is one of the reasons of the keyword argument separation; the details are described in the last section. And `ruby2_keywords` allows you to run the old style even in Ruby 2.7 and 3.0.  As there is no `ruby2_keywords` defined in 2.6 or prior, please use [ruby2_keywords gem](https://rubygems.org/gems/ruby2_keywords) or define it by yourself:
```
jeremyevans: `final section` instead of `last section` here as well, I think.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+
+def foo(*args, **kwargs, &block)
+  target(*args, **kwargs, &block)
+end
+
+foo({})       #=> Ruby 2.7: []   ({} is dropped)
+foo({}, **{}) #=> Ruby 2.7: [{}] (You can pass {} by explicitly passing "no" keywords)
+```
+
+An empty Hash argument is automatically converted and absorbed to `**kwargs`, and the delegation call removes the empty keyword hash, so no argument is passed to `target`.  As far as we know, this is the only one corner case.
```
jeremyevans: `only corner case` instead of `only one corner case`

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+if RUBY_VERSION < "2.7"
+  def accept_and_save_args(*args, &block)
+    @args, @block = args, block
+  end
+
+  def pass_saved_args
+    target(*@args, &@block)
+  end
+else
+  ruby2_keywords def accept_and_save_args(*args, **kwargs, &block)
```
jeremyevans: This is incorrect.  You can't use `ruby2_keywords` on a method that accepts keyword arguments.

I would remove this example of checking Ruby version, since `ruby2_keywords def accept_and_save_args(*args, &block)` will correctly pass positional arguments and keywords to `target`.

I suppose you could remove `ruby2_keywords` from the else branch, but that results in significantly more code, and it is also slower in CRuby.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,377 @@
...
+```
+def foo(**kwargs)
+  kwargs
+end
+foo("key" => 42)
+  #=> Ruby 2.6 or before: ArgumentError: wrong number of arguments
+  #=> Ruby 2.7 or later: {"key"=>42}
+```
+
+If a method accepts both optional and keyword arguments, the Hash object that has both Symbol keys and non-Symbol keys was splitted into two in Ruby 2.6.  In Ruby 2.7, both are accepted as keywords because non-Symbol keys are allowed.
```
jeremyevans: `split in two` instead of `splitted into two`

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: Fixed, thank you!

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: Fixed, thanks!

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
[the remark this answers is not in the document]
mame: Imported the braces, thanks!

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: Good catch!  Changed at 3c2091b2.  I'm unsure if we should keep a unrecommended advice (`$VEBOSE = nil`).

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: Wow, I can't understand why I wrote this.  Removed at 213e70fb.

Do you have any example where a version check is needed?

jeremyevans: I suppose version checking could be needed if you are checking if the last argument is a hash, and using `Hash#merge`:

```ruby
def foo(*args)
  args[-1] = args[-1].merge(bar: false) if args[-1].is_a?(Hash)
  target(*args)
end
```

This wouldn't work with `ruby2_keywords` as the hash returned by `Hash#merge` does not have the flag set on it.  In this case you may want a version check, switching to using keywords on 2.7+.

eregon:
The section about delegation looks good. However I would like to see some mention of when ruby2_keywords will be removed and what code instead should be used to work on 2.6 and after it's removed.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,341 @@
...
+
+Here is one of the hardest cases:
+
+```
+def foo(x=1, **kwargs)
+  p [x, kwargs]
+end
+
+foo({}, **{}) #=> ?
+```
```
eregon: This example is unconvincing for me because it uses `**{}` which almost no real Ruby code uses today. I think a simpler example using e.g. an optional argument and showing both possible desired outcomes would be much better.
I.e., if we did all those changes to save this I'd say not worth it.

jeremyevans:
> The section about delegation looks good. However I would like to see some mention of when ruby2_keywords will be removed and what code can be used to work on 2.6 and after it's removed.

I don't think there has been a decision on that.  There is a tentative idea of after 2.6 EOL.  At that point, we should not concern ourselves with 2.6 compatibility.  We could say that when/if `ruby2_keywords` is removed in the future, you could switch to explicit keyword argument forwarding.

My personal opinion is that `ruby2_keywords` should stay until both of the following are true:

* 2.6 is EOL.
* Explicitly forwarding keyword arguments is not slower than using `ruby2_keywords` in CRuby.

eregon:
> We could say that when/if `ruby2_keywords` is removed in the future, you could switch to explicit keyword argument forwarding.

I would be fine with that.
But currently nothing is said and I feel it's dishonest to show a solution (`ruby2_keywords`) and pretend it will stay forever when it's not been decided yet and likely will not stay forever.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,341 @@
...
+foo({})       #=> Ruby 2.7: []   ({} is dropped)
+foo({}, **{}) #=> Ruby 2.7: [{}] (You can pass {} by explicitly passing "no" keywords)
+```
+
+An empty Hash argument is automatically converted and absorbed to `**kwargs`, and the delegation call removes the empty keyword hash, so no argument is passed to `target`.  As far as we know, this is the only corner case.
+
+As noted in the last line, you can work around this issue by using `**{}`.
+
+If you really worry about the portability, use `ruby2_keywords`.  (Acknowledge that Ruby 2.6 or before themselves have tons of corner cases in keyword arguments. :-)
+
```
eregon:
```suggestion
`ruby2_keywords` might be removed in the future after Ruby 2.6 reaches end-of-life. At that point, we recommend to explicitly delegate keyword arguments (see Ruby 3 code above). If you still need to support Ruby 2.6 at that time, you will need a version check and use a mix of both approaches.

```

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: Thanks.  Well, I'm unsure what you want to do by this code.  It converts the calls:

* `foo(1, 2, 3)` to `bar(1, 2, 3)`
* `foo(1, 2, 3, {})` to `bar(1, 2, 3, {bar: false})`
* `foo(1, 2, 3, **{})` to `bar(1, 2, 3, bar: false)`

Is this really intended?

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,341 @@
...
+foo(k: 1) #=> ArgumentError: unknown keyword k
+```
+
+## Why we're deprecating the automatic conversion
+
+The automatic conversion initially appeared to be a good idea, and worked well in many cases.  However, it had too many corner cases, and we have received many bug reports about the behavior.
+
+Automatic conversion does not work well when a method accepts optional positional arguments and keyword arguments.  Some people expect the last Hash object to be treated as a positional argument, and others expect it to be converted to keyword arguments.
+
+Here is one of the hardest cases:
```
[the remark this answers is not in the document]
mame: I meant 難しい (most difficult to understand and handle).  "most confusing" looks perfect.  Thanks!

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: I agree with the first two sentences.  The last one is to allow users to use 2.6 even after EOL?  It is non-sense to me.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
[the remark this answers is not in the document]
mame: Imported, thanks.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
eregon: > `ruby2_keywords` might be removed in the future after Ruby 2.6 reaches end-of-life.
> At that point, we recommend to explicitly delegate keyword arguments (see Ruby 3 code above).

Third sentence:
> If you still need to support Ruby 2.6 at that time, you will need a version check and use a mix of both approaches.

I mean if some gem wants to support some Ruby 3.x which would not have `ruby2_keywords` *and* Ruby 2.6, they will need a version check like:
```ruby
if RUBY_VERSION >= "3.0" # or if !defined?(ruby2_keywords) if we are careful with the ruby2_keywords gem.
  def foo(*args, **kwargs, &block)
    target(*args, **kwargs, &block)
  end
else
  ruby2_keywords def foo(*args, &block)
    target(*args, &block)
  end
end
```

Many gems support Ruby versions far after the corresponding CRuby version is officially EOL, so we cannot ignore that case.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -0,0 +1,341 @@
...
+
+In most cases, you can avoid the incompatibility by adding the _double splat_ operator. It explicitly specifies passing keyword arguments instead of a `Hash` object. Likewise, you may add braces `{}` to explicitly pass a `Hash` object, instead of keyword arguments. Read the section "Typical cases" below for more details.
+
+In Ruby 3, a method delegating all arguments must explicitly delegate keyword arguments in addition to positional arguments.  If you want to keep the delegation behavior found in Ruby 2.7 and earlier, use `ruby2_keywords`.   See the "Handling argument delegation" section below for more details.
+
+## Typical cases
+
+Here is the most typical case. You can use double splat operator (`**`) to pass keywords instead of a Hash.
+
+```
```
eregon: How about using:
```
```ruby
```
so the code is also syntax-highlighted? I think that would improve readability quite a bit.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: I'm unsure if I meant what I'd like to say... Of course, people can use 2.6 after EOL, at their own risk.  I think it is non-sense for us (Ruby core team) to care and mention such a case.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
jeremyevans: It actually converts `foo(1, 2, 3, **{})` to `bar(1, 2, 3)`, since the double splat of empty hash does not pass an argument.

It would convert `foo(1, 2, 3, a: 'b')` to `bar(1, 2, 3, {a: 'b', bar: false})` (even with `ruby2_keywords`.  Under Ruby <2.7, the expectation is probably it would modify keywords and result in `bar(1, 2, 3, a: 'b', bar: false)`, but that doesn't happen as `Hash#merge` returns a hash without the keyword flag.  This is a possible case where you would want to use different method definitions, with explicit keywords on Ruby 2.7+ and the existing code on Ruby 2.6:

```ruby
if RUBY_VERSION < '2.7'
  def foo(*args)
    args[-1] = args[-1].merge(bar: false) if args[-1].is_a?(Hash)
    target(*args)
  end
else
  def foo(*args, **kw)
    target(*args, **kw, bar: false)
  end
end
```

Note that this does not result in the exact same behavior. In 2.6 this will modify a positional hash argument, while the 2.7 version does not.  Also, if `target` doesn't accept keywords, the 2.7+ code passes an additional `{bar: false}` argument to it.

This isn't a good example either, as the code given for 2.7 works as far back as 2.0.  So you should probably just switch to using keywords unless you need to support Ruby <2.0. 

So I can't think of a reasonable case where you would want to use a version check.  We should probably just not mention version checking.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
eregon: I think we need to care a bit about that case, because the kwargs changes are going to be pretty ugly for those cases. I think it's better to write something than nothing which might be interpreted as "the Ruby core team doesn't care about that case and their only written solution is to drop Ruby 2.6 support, which might not be an option for that gem".

Could you approve the suggestion and then adapt the third sentence in a next commit?

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: Actually, it is not just an artificial example.  I've changed the explanation to a more descriptive one in 874673ad.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
```diff
@@ -308,14 +308,23 @@ Automatic conversion does not work well when a method accepts optional positiona
...
+end
+
+foo({}) => [{}, {}]
+bar({}) => [1, {}]
+
+bar({}, **{}) => expected: [{}, {}], actual: [1, {}]
 ```
 
-While one might intuitively expect the above code to print `[{}, {}]`, it in fact prints `[1, {}]` in Ruby 2.6.  This is because `**{}` is ignored by the parser in Ruby 2.6, and the first argument `{}` is automatically converted to keywords (`**kwargs`).
+In Ruby 2, `foo({})` passes an empty hash as a normal argument (i.e., `{}` is assigned to `x`), while `bar({})` passes a keyword argument (i.e, `{}` is assigned to `kwargs`).  So `any_method({})` is very ambiguous.
```
eregon: Much more convincing :+1:

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
jeremyevans: I don't think we want to mention version checks as a way to keep far future and 2.6 compatibility.

* Every place you are mentioning a version check will result in slower code in CRuby 3.0 (unless someone rewrites CRuby's keyword support). 
* We don't know what the status will be with `ruby2_keywords`, but it is likely it will be present until at least 2.6 EOL, which means Ruby 3.2 will still support it.  We can reevaluate later whether it should be kept longer.
* We shouldn't go out of our way to cater to people who use EOL versions.
* If we do drop `ruby2_keywords`, we can have a new (and much smaller) document that shows users how to cope with that change.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: Annotations are added.  61fd17bb

Unfortunately, syntax-highlighting seems not to work currently.
https://www.ruby-lang.org/en/documentation/faq/1/

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: > Could you approve the suggestion and then adapt the third sentence in a next commit?

Done, and removed the third sentence.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: I understand, jekyll requires `{% highlight ruby %}...{% endhighlight %}`.  Will fix soon

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
mame: Thanks!

> So I can't think of a reasonable case where you would want to use a version check. We should probably just not mention version checking.

Agreed.  Let's add a reasonable case if/after we encounter it.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
eregon: > Every place you are mentioning a version check will result in slower code in CRuby 3.0 (unless someone rewrites CRuby's keyword support).

That depends on the check. If it was something like `if RUBY_VERSION > "3.2"` or `if !defined?(ruby2_keywords)` then it wouldn't slow down code for any of those version before `ruby2_keywords` is removed (that needs some care in the `ruby2_keywords` gem to only define the `ruby2_keywords` method if < 2.7, it would anyway be incorrect for later versions, I'll make a PR).
After `ruby2_keywords` is removed it can't be used obviously, and I'd say the earlier the better since we'll have to face it eventually anyway.

I'm fine to not have code about the version check in the blog post.
Just the third sentence above is better than nothing.

FWIW, I still consider it's a huge hack we need a magic `ruby2_keywords` modifier to get delegation working in a programming language, and that we broke `*args`-delegation. The blog post already mentions:
```
def foo(*args, **kwargs, &block)
  target(*args, **kwargs, &block)
end
```
as it should, because that is the logical/clean/intuitive way to do delegation in Ruby 3+.
(I'm also fine with new ways like `...`)

IMHO we should focus on making those fast, because those ways are clean and intuitive.
`ruby2_keywords` won't be forever, I think it makes very little sense to use `ruby2_keywords` in new code after the Ruby 3.0 release and IMHO a hack should not be used just for performance of delegation.

Back to this blog post, from what I see you're not against my original suggestion, but rather about the specifics of the code above. So can we just apply my suggestion and propose further tweaks if desired?

mame:
Thank you for your review and comments!  I really appreciate you all.

The commit log in this PR seems too long to make CI fail :-)  I'll squash the commits, and merge it soon.

Even after this article is published, we can modify it with history.

en/news/_posts/2019-12-06-separation-of-positional-and-keyword-arguments-in-ruby-3-0.md
eregon: > Done, and removed the third sentence.

OK, better than nothing, thanks.

PR #2316, opened 2019-12-18
Fix description about frozen string

JunichiIto:
As far as I know, they reverted only `Symbol#to_s` for the frozen string feature.

Here is the result against 2.7.0-rc1:

```
irb(main):001:0> :a.to_s.frozen?
=> false
irb(main):003:0> true.to_s.frozen?
=> true
irb(main):004:0> false.to_s.frozen?
=> true
irb(main):005:0> nil.to_s.frozen?
=> true
irb(main):006:0> Enumerable.name.frozen?
=> true
```

See also:

- https://bugs.ruby-lang.org/issues/16150
- https://bugs.ruby-lang.org/projects/ruby-trunk/repository/git/revisions/bea322a352d820007dd4e6cab88af5de01854736/diff


nurse:
Thank you for catch up!

PR #2318, opened 2019-12-19
Switch off autoplay in 2.7.0-rc1 post

stomar:
It's more annoying than helping. At the time when one reaches the paragraph, the video has been playing for some time already, and one needs to stop and rewind. It's better to simply start the video at the time one wants to see it.


stomar:
@nurse Some feedback would be welcome, especially in regard to the upcoming final release.

ismailarilik:
I think it is better to make this kind of language-agnostic changes to all translations since it is faster to make them on one branch than making them on different branchs via different PRs.

ismailarilik:
Sorry about the comment above, I have thought that there was corresponding Turkish translation but I have realized that it wasn't merged yet.

nurse:
I think it should be autoplay because the video is just a decoration and readers feel the experience.

stomar:
ACK; fixed.

PR #2345, opened 2020-01-02
Follow 2344 (tr)

ismailarilik:


stomar:
I cannot review `tr`; I only would like to note that the change to the `en` version was language related, more a question of style, and might not necessarily apply to other languages. When the previous version sounds right in `tr`, it‘s not necessary to change it.

PR #2351, opened 2020-01-12
Translation of release of 2.7.0-preview3 (es)

vtamara:


vtamara:
@dabit, @vic, @soveran, @gfvcastro could you please review?

PR #2163, opened 2019-10-09
Add DEV link to blog page (en)

benhalpern:
This pull request adds a link for the [Ruby tag on DEV](https://dev.to/t/ruby). DEV is the fastest growing community blogging ecosystem in software and makes a lot of sense right here in this list.

There are new posts consistently flowing into the tag, and it is becoming one of the best Ruby resources available.

Issue #1894, opened 2018-11-28
Follow up changes of upstream (ko)


yous:
@ruby/www-ruby-lang-org-i18n-ko We can keep track of the following list of pull requests or commits on upstream.

- [ ] #1836 
- [ ] #1877 
- [ ] #1888

yous:
#1879 applies changes of #1877.

yous:
#1909

riseshia:
#1900 -> We don't have translated faq page XD

yous:
Okay, then I'll fill out the missing files.

yous:
Just updated files that are missing in ko.

yous:
#1920

yous:
#1926

yous:
Removed `/en/downloads/mirrors/index.md` as it's not supposed to be translated. Others are saying "in development", so I commented next to the links.

riseshia:
`/en/conduct/index.md` seems to be unreachable (updated ^)

yous:
Isn't it available? https://www.ruby-lang.org/en/conduct/

riseshia:
I couldn't find inner link from this repository...
but It is searchable state... 🤔 
https://www.google.com/search?q=ruby-lang+conduct&oq=ruby-lang+conduct

yous:
I couldn't find a link in this repository neither. But I found <https://bugs.ruby-lang.org/issues/12004>, and #1297. Thanks for checking the list!

yous:
#1953

yous:
#1963, #1965, #1966

riseshia:
#1965, #1966 fixed at #1968, #1969, so #1963 remained.

yous:
#1682

yous:
#1681

yous:
<https://github.com/ruby/www.ruby-lang.org/commit/e2f18e4bb2204d32f73b78412be0e8ff16627253> Why was this commit pushed directly instead of through pull requests? This kind of changes are hard to track, even though I'm watching this repository.

yous:
#1466

yous:
https://github.com/ruby/www.ruby-lang.org/commit/da7c5408a5a6fb2d5c79555cb28700c096d578ba

yous:
#2002 Note that 2019-03-05-multiple-vulnerabilities-in-rubygems.md is also changed.

yous:
#2003

yous:
[...]

PR #2354, opened 2020-01-16
Translate "DEV link to blog page" #2163 (ko)

riseshia:
Master issue: #1894
Translated PR: #2163 

Actual diff is ca88db4
Could you review this?


riseshia:
Thanks~

stomar:
When squashing commits, please fix the commit messages in a meaningful way; the default of listing all individual commit messages of intermediate small steps like below often doesn't make sense in the final history, but only adds noise (the summary in the first line would have been enough).

```
Translate "DEV link to blog page" #2163 (ko) (#2354)

* backport "DEV link to blog page" #2354
* Translate text
```

riseshia:
I will be more careful..! Thanks for the comment.

PR #2357, opened 2020-01-17
Fix typos on 2.7.0 releases (ko)

yous:
This is a follow-up of #2356.

Sorry for the late review.


yous:
Still waiting the builds.

PR #2363, opened 2020-02-02
Add posts about Rubyist Magazine 61 published (ja)

miyohide:
RubiMa Editors have released Rubyist Magazine 61 (in japanese).


hsbt:
@miyohide Thanks always!

PR #2365, opened 2020-02-03
Add HTML anchors for the article of Ruby 3's keyword argument separation

mame:
to make it useful to refer


ismailarilik:
This will not be translated, right?

JuanitoFatas:
> This will not be translated, right?

Yes we don‘t need to translate the anchor. Could add to translations so they also have anchors for people to reference.

ismailarilik:
> Yes we don‘t need to translate the anchor.

For this kind of PRs, isn't it efficient to patch the changes to all available translations?

PR #2384, opened 2020-03-10
Fix typo in english libraries index page

markim:
"As most programming languages"

to

"As **with** most programming languages"

Perhaps the original way is correct, but grammatically it did not make sense to me. Have a great day!


JuanitoFatas:
Thanks @markim and @larouxn for review.

PR #2392, opened 2020-03-26
Translate "reverted" word in 2.7.0-preview3 (ru)

aleksandrilyin:
Translate word.


lex111:
@aleksandrilyin thanks!

PR #2393, opened 2020-03-26
Fix syntax highlighted code blocks in lists (ru)

aleksandrilyin:
Refer to commit [6494161](https://github.com/ruby/www.ruby-lang.org/commit/6494161bd6304e357b49bdf220ee01fddc2777b1).


lex111:
Спасибо!

PR #2396, opened 2020-03-26
Add rubybib.org link in sidebar (ru)

aleksandrilyin:
Add rubybib.org link.

### DIFF
--- a/_data/locales/ru.yml
+++ b/_data/locales/ru.yml
@@ -44,7 +44,7 @@ sidebar:
     # books:
     #   text: Книги
     rubybib:
-      text: Академическое исследование
+      text: Научные публикации
       url: https://rubybib.org/
     libraries:
       text: Библиотеки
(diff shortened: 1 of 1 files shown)


_data/locales/ru.yml
```diff
@@ -43,6 +43,9 @@ sidebar:
       url: /ru/documentation/
     # books:
     #   text: Книги
+    rubybib:
+      text: Академическое исследование
+      url: https://rubybib.org/
```
arbox: @aleksandrilyin , I don't think, it is a sutable description for this point :) Something like "Научные публикации", "Академическая библиография" or at least "Академические исследования" would fit better (in my opinion).

aleksandrilyin: @arbox thank you for review! "Научные публикации" looks good and accessibly.
@lex111 what do you think about translation for "Academic Research"?

lex111:
Мне больше нравится вариант "научные работы", хотя "публикации" тоже хорошо, спасибо @aleksandrilyin и @arbox!

PR #2410, opened 2020-04-01
Move Ruby 2.4 to EOL

eregon:
https://www.ruby-lang.org/en/news/2020/03/31/ruby-2-4-10-released/

cc @mame @unak

### DIFF
--- a/_data/branches.yml
+++ b/_data/branches.yml
@@ -19,12 +19,12 @@
   eol_date:
 
 - name: 2.5
-  status: normal maintenance
+  status: security maintenance
   date: 2017-12-25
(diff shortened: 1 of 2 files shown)


eregon:
Could we merge this?
Should we move 2.5.8 to `security_maintenance`, or do that later?

hsbt:
We will update them after publishing the official annoucement by @unak

eregon:
I'm fine to close this if we're going with the more complete https://github.com/ruby/www.ruby-lang.org/pull/2412

eregon:
I updated this, I think together with #2420 it's complete.

PR #2411, opened 2020-04-02
update link to Poignant Guide

stadtwald:
The primary website for the Why's (Poignant) Guide to Ruby appears to have moved.  The old link now appears to be broken, so this patch updates the links in the documentation.


vtamara:
ok for es

PR #2415, opened 2020-04-04
Fix CVE-2020-10663 name in latest news posts (en, tr)

aleksandrilyin:
Typo.


JuanitoFatas:
Good catch! Thanks 👍

darix:
when will this be deployed?

aleksandrilyin:
@darix I think deploy will be later because he is not automatic in the production (read about Deployment at this [link](https://github.com/ruby/www.ruby-lang.org/wiki/Deployment)).

PR #2424, opened 2020-04-08
Add rubybib to sidebar (ko)

riseshia:
which is from https://github.com/ruby/www.ruby-lang.org/blob/c83b6eee831b847c39aa0f0f76397a6d0a2eb9be/_data/locales/en.yml#L47-L49


riseshia:
cc @aleksandrilyin

riseshia:
Thanks!! :)

PR #2430, opened 2020-04-14
Translate Ruby 2.4 support ended post (ru)

aleksandrilyin:
cc @lex111


lex111:
@aleksandrilyin спасибо!

PR #2449, opened 2020-05-10
Remove "*" symbol in "Separation of positional and keyword arguments in Ruby 3.0" post (id)

aleksandrilyin:
/cc @meisyal
@meisyal maybe you wanted to make italics, but it didn't work in these places. Can we remove this symbol?


meisyal:
Sure @aleksandrilyin. I just realized if that didn't work.

meisyal:
Merged. Thank you @aleksandrilyin

PR #2460, opened 2020-06-26
Remove Google Search; Google Analytics

resynth1943:
## Remove Google Search

This commit replaces the Google search box to a DuckDuckGo search box.

I have chosen not to use their search box, but instead I have incorporated
this functionality into the search box that was previously being used.

I have also added short codes to every locale file, as seen in the previous
commit. DuckDuckGo supports localisation, so this should make the search
experience nicer for visitors of the Ruby website.

I am unfamiliar with Google, so this may be a guess, but I have witnessed
that some CSE special codes were missing from the locale files. I can only
guess that these were previously being used for localisation, e.g. choosing
a specific language for the search box.

If so, this should be a massive win for users of this site that are: not fluent
in the English language; do not have a Google account tailored to their language.

Moreover, this change will be a brilliant win for the privacy of the users of
your site. As Google is an advertising company, we can only guess what they do
with the data they collect on millions of users every day.

In my personal opinion, tackling this abhorrent abuse of data should also be 
peformed by FOSS projects. Just think: every time someone searches, you've just
saved their online privacy! :D

Theoretically, DuckDuckGo should in fact be more performant than a Google search,
too. I have friends who have just confirmed this, which speaks volumes of
DuckDuckGo as a service.

---

## Remove Google Analytics

This commit removes the Google Analytics 'functionality' from this page. Following on
from the previous commit, this change also greatly improves the privacy of 
unsuspecting users.

If you must have an analytics system, might I suggest either:
- Plausible: https://plausible.io (offers a 50% discount for FOSS projects)
- Goatcounter: https://goatcounter.com
- Matomo: https://matomo.org
- Fathom: https://usefathom.org

These solutions are all well-tested and open-source. This, of course, means they
can be independently audited for privacy and security. Instead of using a proprietary
nonfree system such as Google Analytics, perhaps we could work together to replace
it with something slightly more ethical?

Furthermore, the majority of privacy features in evergreen browsers *block* Google
Analytics from being run. I can not say the same for solutions such as Plausible,
which suggests they are trusted for their privacy by independent persons.

Thankfully, this change is relatively easy to make, as you have a brilliant
taste in build tools! Jekyll is, thankfully, rather easy to work with.

If you need any more information, please contact me!


hsbt:
It's good to migrate the internal search engine like Elasticsearch from company service. But I'm not sure what advance DuckDuckGo from Google CSE.

resynth1943:
> It's good to migrate the internal search engine like Elasticsearch from company service. But I'm not sure what advance DuckDuckGo from Google CSE.

I mean, we could try a different search engine.

---

I'd like to talk more about your ElasticSearch idea. Would you be open to reopening this, so I could implement it if poss?

resynth1943:
Hey @hsbt, would you be happy for me to open an issue so we could discuss your Elasticsearch idea further?

hsbt:
I also maintain the infrastructure of *.ruby-lang.org. I have no time to handle it. I understood you have a concern to use Google product. But I don't have it. So, This issue is low priority for me.

PR #2477, opened 2020-09-14
Fix double rendering of news archives

ashmaroli:
Since the generated archives are pushed into the `site.pages` array, they will be eventually *rendered* and *written* to destination like regular pages even if the generator has `priority: low`. So manually having those objects call `:render` and `:write` is redundant and contributes to increased build times.

P.S. While I did check if this change caused any difference in the files ultimately written to `./_site`, do let me know if I missed a negative side-effect from this change.

/cc @stomar

### DIFF
--- a/_plugins/news.rb
+++ b/_plugins/news.rb
@@ -151,8 +151,6 @@ module Jekyll
           years.values.map(&:values).flatten
         )
 
-        index.render(site.layouts, site.site_payload)
-        index.write(site.dest)
         site.pages << index
 
         years.each do |year, months|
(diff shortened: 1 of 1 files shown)


ashmaroli:
Pinging @nagachika for alternate set of :eyes:

hsbt:
What difference before and after for users?

ashmaroli:
> difference before and after for users?

For visitors to `www.ruby-lang.org`,  none.
For contributors and developers of the site, faster builds..

stomar:
I confirmed the generated `_site` is unchanged (`diff -r`).

stomar:
Thanks @ashmaroli, a good catch!

And sorry for the late response.

stomar:
Interestingly, I was reminded about your PR when working on a problem in `_plugins/news.rb`. It turned out to be related to a commit you made to Jekyll... funny coincidence :smiley:.

See #2591.

PR #2505, opened 2020-10-14
Fix italic text

riseshia:
https://www.ruby-lang.org/ko/news/2020/09/25/ruby-3-0-0-preview1-released/

`_인터페이스 타입_을` isn't parsed 😇

PR #2506, opened 2020-10-15
Fix italic text for Korean

yous:
This is a follow-up of #2505.

Using `*` instead of `_` works even if there are no surrounding spaces: [example](https://github.com/ruby/www.ruby-lang.org/blame/master/ko/documentation/ruby-from-other-languages/index.md#L17) ([web](https://www.ruby-lang.org/ko/documentation/ruby-from-other-languages/)).


riseshia (approved):
thanks!!!!

PR #2512, opened 2020-10-20
Fix mixed content error on Fukuoka Ruby Award

VagishVela:
The image linked is using http instead of https which is causing a mixed content error, which now results in a "Content not secure" near the padlock on modern browsers.

I noticed this in multiple places so fixed them all.


JuanitoFatas (approved):
Thanks!

PR #2513, opened 2020-10-20
Remove broken europe.railsconf.org links

VagishVela:
The URL http://europe.railsconf.org is no long there, and an alternative is not easily accessible. It looks like the global conferences are in all on spot this year as RailsConf Coach Edition.

RailsConf Europe is good to mention historically (most of the pages have context about the conference too), so I removed the link and left the mention.


yous:
`it` and `tr` files still have `[RailsConf Europe][13]` and `[RailsConf Avrupa][13]`, respectively.

VagishVela:
> `it` and `tr` files still have `[RailsConf Europe][13]` and `[RailsConf Avrupa][13]`, respectively.

Good catch! I've created a PR to fix that. Thanks :)

PR #2521, opened 2020-10-22
Fix build

JuanitoFatas:
- Remove non-existent architecture keyword
- Upgrade actions/checkout to v2
- Use [cache dependencies automatically](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically) by ruby/setup-ruby
- Remove CUSTOM_RUBY_VERSION 
- No need to install Bundler

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,28 +18,16 @@ jobs:
       - name: Dump environment
         run: env | sort
       - name: Checkout ruby/www.ruby-lang.org
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
           fetch-depth: 1
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
-          architecture: x64
+          bundler-cache: true
       - name: Dump Ruby version
         run: ruby -v
-      # override the Ruby version specified in the Gemfile
-      - name: Set CUSTOM_RUBY_VERSION for Gemfile
-        run: echo ::set-env name=CUSTOM_RUBY_VERSION::$(ruby -e 'puts RUBY_VERSION')
-      - name: Dump CUSTOM_RUBY_VERSION
-        run: echo $CUSTOM_RUBY_VERSION
-      - name: Cache gem bundle
-        uses: actions/cache@v1
-        with:
-          path: vendor/bundle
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -18,28 +18,16 @@ jobs:
...
-      - name: Dump CUSTOM_RUBY_VERSION
-        run: echo $CUSTOM_RUBY_VERSION
-      - name: Cache gem bundle
-        uses: actions/cache@v1
-        with:
-          path: vendor/bundle
-          key: ${{ runner.os }}-${{ matrix.ruby }}-bundle-${{ hashFiles('Gemfile.lock') }}
-      - name: Install bundler
-        run: gem install bundler --no-document
       - name: Install gem bundle
```
yous, line 31: It seems that `bundle install` will be ran automatically by `ruby/setup-ruby`: https://github.com/ruby/setup-ruby#caching-bundle-install-automatically

JuanitoFatas: Hmm I think we still need to run `bundle install` given https://github.com/ruby/setup-ruby/blob/a195098f6b1e9074390215a49c1c14e82151a289/.github/workflows/test.yml#L74

yous: They're also running `bundle install` twice I think. 😅

That line is added [8 months ago](https://github.com/ruby/setup-ruby/blame/a195098f6b1e9074390215a49c1c14e82151a289/.github/workflows/test.yml#L74), but `bundle-cache: true` is added [5 months ago](https://github.com/ruby/setup-ruby/blame/a195098f6b1e9074390215a49c1c14e82151a289/.github/workflows/test.yml#L42). Seeing this commit, it runs `bundle install` automatically. https://github.com/ruby/setup-ruby/commit/d57231c9a394918decfec234a36c9c79116254b7#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R56

On "Checks" tab, clicking "Setup Ruby" shows the following:
```
Run ruby/setup-ruby@v1
Downloading Ruby
Extracting Ruby
Installing Bundler
bundle install
```
Clicking "Install gem bundle" shows every entry with `Using ...`.

JuanitoFatas: Oh you’re absolutely right. Thanks for detailed explanations. I will add a follow-up PR.

eregon: Yes, `gem install bundler & bundle install` is not necessary with `bundler-cache: true`.
I'll update that test file or add a comment.

PR #2522, opened 2020-10-22
We dont need to run bundle install

JuanitoFatas:
it is already performed by ruby/setup-ruby. See https://github.com/ruby/www.ruby-lang.org/pull/2521#discussion_r510093800.


yous (approved):
Thanks!

Issue #2433, opened 2020-04-18
The link to xmp is broken


ngzhio:
https://github.com/ruby/www.ruby-lang.org/blob/dfa7e9d5249f065ef4cb3b1ae96772bfd54721e2/en/documentation/faq/1/index.md#L294

hsbt:
Can you create the pull request to remove them?

ngzhio:
> Can you create the pull request to remove them?

Remove them? Isn't there any new address of `xmp`?

hsbt:
Feel free to choose them. I have no opinion for it.

siaw23:
I think this should be closed as merged from https://github.com/ruby/www.ruby-lang.org/pull/2523 @hsbt @yous

PR #2523, opened 2020-10-22
Remove RAA XMP link as no longer exists

VagishVela:
The RAA is gone, and the location for the link is also gone.

This closes the issue #2433


JuanitoFatas (approved):
👍🏼

PR #2536, opened 2020-10-26
Fix syntax highlighted code blocks in lists (en)

stomar:
Use fenced code blocks instead of Jekyll's `highlight` tag.

The latter does not work in list items: the highlight tag terminates the list, so that the highlighted code is not part of the list item; subsequent list items start a new list.

Visually, this fixes wrong indentation of the code blocks, which had less indentation than the surrounding list items.

_Before this fix (note the missing indentation and the strange padding below the code block):_

code_blocks


stomar:
Generally, **GitHub style fenced code blocks** should be used now (which didn't work some time ago, but now do).

PR #2540, opened 2020-10-27
Stop using Travis CI

stomar:
Travis CI seems to be very slow. It takes a really long time for jobs to get started (at least for me in the last days).

Is it still needed?

(Possibly for notifications?)


JuanitoFatas (approved):
I also share the same sentiment 👍🏼

hsbt:
I forgot to keep Travis CI when we start to use GH Actions. If we found some issues, we can revert this easily.

stomar:
Anything else that would have to be adjusted, e.g. in the repository's settings etc.?

PR #2553, opened 2020-11-20
Remove CI tests for Ruby 2.6

stomar:
Tests for 2.6 fail since the Ruby version was locked to 2.7.2
in the Gemfile, see a96ca92fcb622f450dda1c385a5f01aa86ee35d4.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby: [2.6, 2.7]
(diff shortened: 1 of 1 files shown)


stomar:
cc @hsbt

For now, simply removing the tests seems fine. But for future updates it might be nice to have tests for multiple versions in parallel, which does not work with the current Gemfile.

hsbt:
Thanks for fixing this.

We need to declare `ruby` syntax for heroku. This change is reasonable for it.

PR #2563, opened 2020-12-08
Remove duplicated "Find pattern is added."

benoittgt:
Hello

This part is duplicated. It can be seen just earlier https://github.com/ruby/www.ruby-lang.org/compare/master...benoittgt:benoittgt-remove-dup-find-pattern?expand=1#diff-c6345d79a5e1d212e019adf651677f24c74206998b20198fa191701b2052e703R161-R169

Thanks for the release and the great changelog.

🎉


JuanitoFatas (approved):
Thanks!

PR #2580, opened 2020-12-25
Add meta image and description

yuki24:
So that services like Slack can extract the summary of the page and show it to the user.


JuanitoFatas (approved):
☺️

PR #2583, opened 2020-12-25
Fix vers of standard libraries in 3.0.0 release note in en and ja, and fix typo `Pysch'

scivola:


JuanitoFatas (approved):
Thank you!! 👍🏼

PR #2588, opened 2020-12-25
Restrict Jekyll to < 4.2.0 and update gem bundle

stomar:
Using Jekyll 4.2.0 breaks the sidebars on the yearly and monthly news archive pages: they are the same as on the news/ page, listing "Archives by Year" (without any entries) and "Syndicate".


stomar:
The rendered _site does not change with this bundle update.

PR #2591, opened 2020-12-27
Fix generated news archive pages for Jekyll 4.2.0 and update bundle

stomar:
Before this change, with Jekyll 4.2.0 all news archive pages used `news.html` as layout, also the yearly and monthly archive pages, leading to wrong sidebars that list "Archives by Year" (without any entries) and "Syndicate".

The reason is that `Page#relative_path` is a memoized value and Jekyll 4.2.0 changes when `relative_path` is first called during generation of the archive pages.

Other than before it is now called from `read_yaml`, see

https://github.com/jekyll/jekyll/commit/26086409bff1d9f903ef723c6d4225c519c01e5c

`YearlyArchive` and `MonthlyArchive` call `ArchivePage#new` (via `super`), which uses `read_yaml`, so that from then on `relative_path` is set and will not change anymore.  Therefore the reassignments of `@dir` in YearlyArchive and MonthlyArchive after the `super` calls have no effect on `relative_path`. Hence *all* archive pages have their relative path set to `<lang>/news/index.html`.  Then for rendering, apparently the content of the page under the relative path is used (i.e. the `news.html` layout) instead of the in fact correctly set `content` of the respective archive page (i.e. `news_archive_year.html` or `news_archive_month.html`).

This issue is fixed by not changing `@dir` after calling `super`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 ruby "~> 2.7.2"
 
 gem "rake"
-gem "jekyll", "~> 4.0", "< 4.2.0"
+gem "jekyll", "~> 4.0"
 gem "rouge"
 
 gem "unicorn"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -15,21 +15,21 @@ GEM
     http_parser.rb (0.6.0)
     i18n (1.8.5)
       concurrent-ruby (~> 1.0)
-    jekyll (4.1.1)
+    jekyll (4.2.0)
       addressable (~> 2.4)
       colorator (~> 1.0)
       em-websocket (~> 0.5)
       i18n (~> 1.0)
       jekyll-sass-converter (~> 2.0)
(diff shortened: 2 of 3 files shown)


stomar:
I confirmed the generated `_site` does not change (`diff -r`).

ashmaroli:
Uh oh! Looks like I introduced an unintentional change in behavior with Jekyll 4.2, @stomar.
Sorry about breaking the SemVer contract and forcing you to change the signature of your constructors.

Looking into maintaining backwards compatibility now..

stomar:
@ashmaroli

I'm not sure on which side the problem was; possibly reassigning `@dir` hasn't been part of the intended contract.

ashmaroli:
> possibly reassigning `@dir` hasn't been part of the intended contract.

Can't say for sure. But the `Jekyll::Page` class does have `attr_writer :dir` defined; So...
Either ways, I've proposed #2593 to do away with calling `read_yaml` altogether..

PR #2593, opened 2020-12-28
Read archive page layouts only once

ashmaroli:
Currently when building with the `master` branch, **the same layout file** for each `News::ArchivePage` descendant is read multiple times (for every instance):
https://github.com/ruby/www.ruby-lang.org/blob/6b25ddfd2fb2b0320399eb6c3ac1dad7b7260be0/_plugins/news.rb#L28

This set of IO operations can be done away with because Jekyll's layout-reader already takes care of reading layout files.

*NOTE: This change changes the signature of the constructor methods and additionally renders the `LAYOUT` constant in each subclass redundant.*

/cc @stomar

### DIFF
--- a/_plugins/news.rb
+++ b/_plugins/news.rb
@@ -6,7 +6,7 @@ module Jekyll
   module News
     class ArchivePage < Page
 
-      def initialize(site, base, subdir, layout, lang, posts)
+      def initialize(site, base, subdir, lang, posts)
         @site = site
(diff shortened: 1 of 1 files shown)


stomar:
Merged via 46b224e0acd799354bb65eaf9d90fc94ccb062ef.

stomar:
Thanks!

PR #2599, opened 2020-12-30
Remove redundant attribute from archive instances

ashmaroli:
The `@base` attribute is only ever used (in the parent class `Jekyll::Page`) to compute the absolute path, which in turn is of no value for the generated archive pages.
Therefore, simply the constructor signatures by doing away with the redundant parameter.
Additionally, since the `site` instance is already being passed to the constructor, the `base` can be applied easily should a need for it arise in the future.

/cc @stomar

### DIFF
--- a/_plugins/news.rb
+++ b/_plugins/news.rb
@@ -8,9 +8,8 @@ module Jekyll
 
       attr_reader :lang
 
-      def initialize(site, base, subdir, lang, posts)
+      def initialize(site, subdir, lang, posts)
         @site = site
-        @base = base
         @dir  = if subdir
                   File.join(lang, "news", subdir)
                 else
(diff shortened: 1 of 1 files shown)


stomar:
I'm not sure about this change.

I would rather not dive too deeply into resp. depend on implementation details of Jekyll. I would tend to consider the [documentation for generators](https://jekyllrb.com/docs/plugins/generators/) as the public API I am supposed to rely on.

The example code there:

``` ruby
class CategoryPage < Page
  def initialize(site, base, dir, category)
    @site = site
    @base = base
    @dir  = dir
    @name = 'index.html'

    self.process(@name)
    self.read_yaml(File.join(base, '_layouts'), 'category_index.html')
    self.data['category'] = category

    # ...
  end
end
```

ashmaroli:
I see..
The example in the documentation was meant to be just that: *an example*.
But looks like the intention is not conveyed accordingly.

I'll submit a pull request to edit / update the example in the docs upstream.
Thank you for bringing this to my notice.

stomar:
Probably my above statement was a bit too strong; I am aware it is an example.

My argument is that the maintainers here are not experts on Jekyll internals. So I would rather keep it as simple and straightforward as possible, i.e. close to easily accessible resources like the example an jekyllrb.com.

stomar:
I would keep this open for second opinions.

ashmaroli:
No issues, @stomar.
The proposed change isn't critical. So there's no compulsion.

ashmaroli:
@stomar The Jekyll documentation on generators have been updated.
Do let me know what you think of the updated copy from the perspective of a novice Jekyll plugin developer.

PR #2600, opened 2020-12-30
Set basename and ext attrs of archives directly

ashmaroli:
Since the `@name` attribute for all generated archive pages are the same `"index.html"`, the `@basename` and `@ext` attributes are also technically the same values for all instances.
Therefore, set those attributes directly instead of computing strings via series of method calls.

/cc @stomar

### DIFF
--- a/_plugins/news.rb
+++ b/_plugins/news.rb
@@ -16,11 +16,12 @@ module Jekyll
                 else
                   File.join(lang, "news")
                 end
-        @name = "index.html"
+
+        @basename = "index"
+        @ext      = ".html"
+        @name     = "index.html"
 
         @lang = lang
-
-        process(@name)
         @data ||= {}
 
         data["lang"]  = lang


stomar:
I'm not sure about this change.

I would rather not dive too deeply into resp. depend on implementation details of Jekyll. I would tend to consider the [documentation for generators](https://jekyllrb.com/docs/plugins/generators/) as the public API I am supposed to rely on.

The example code there:

``` ruby
class CategoryPage < Page
  def initialize(site, base, dir, category)
    @site = site
    @base = base
    @dir  = dir
    @name = 'index.html'

    self.process(@name)
    self.read_yaml(File.join(base, '_layouts'), 'category_index.html')
    self.data['category'] = category

    # ...
  end
end
```

ashmaroli:
I see..
The example in the documentation was meant to be just that: *an example*.
But looks like the intention is not conveyed accordingly.

I'll submit a pull request to edit / update the example in the docs upstream.
Thank you for bringing this to my notice.

PR #2605, opened 2021-01-01
Improve plugin for news archives

stomar:


stomar:
This doesn't change the generated `_site` (checked with `diff -r`).

PR #2617, opened 2021-02-14
Remove redundant expression in ja quickstart

tanaken0515:
I felt that the expression "Rubyに" was redundant.


hsbt:
Good catch, Thanks!

PR #2631, opened 2021-04-05
Remove macos-latest from GitHub Actions

mame:
It is relatively slow comparing to Ubuntu, which blocks update of the
site.
I don't think it is reasonable to run the test on CI.
Instead, a rapid update is important, especially, for security release.


JuanitoFatas (approved):
👍🏼

PR #2634, opened 2021-04-06
Remove EOLed snapshot-ruby_2_5 and update version of nightly_snapshot

znz:


hsbt:
Thanks, Fixed at https://github.com/ruby/www.ruby-lang.org/commit/4952e4430f85798b614402dad8fa3a837ba06487

PR #2654, opened 2021-05-24
Add an advice for bundler users

mame:


mame:
For the record: @aycabta approved this PR in personal communication

PR #2659, opened 2021-06-10
Remove reference to book list as link is broken

anphase:
The link to the book list in the quickstart guide (http://books.ruby-doc.com/bookstore/) no longer works.


hsbt (approved):

ismailarilik:
This change seems language independent so isn't it more efficient to change all related language files rather than waiting translators to follow.

lex111 (approved):

lex111:
@anphase thanks!
@ismailarilik translations should be fixed in #2660

PR #2690, opened 2021-09-29
Remove abuse enabling language

jacobherrington:
My rationale for these changes are documented on Twitter:
https://twitter.com/JakeHerrington/status/1443286841533374469

### DIFF
--- a/en/community/mailing-lists/ruby-talk-guidelines/index.md
+++ b/en/community/mailing-lists/ruby-talk-guidelines/index.md
@@ -9,9 +9,7 @@ You should follow these guidelines when posting to the ruby-talk mailing list.
 
 
 1. **Always** be friendly, considerate, tactful, and tasteful. We want to
-   keep this list hospitable to the growing ranks of newbies, very
-   young people, and their teachers, as well as cater to fire breathing
-   wizards. :-)
+   keep this list hospitable and safe for everyone.
 
 2. Keep your content relevant and easy to follow. Try to keep your
    content brief and to the point, but also try to include all relevant
--- a/en/conduct/index.md
+++ b/en/conduct/index.md
@@ -18,5 +18,5 @@ commit comments, etc.).
 
  * Participants will be tolerant of opposing views.
  * Participants must ensure that their language and actions are free of personal attacks and disparaging personal remarks.
- * When interpreting the words and actions of others, participants should always assume good intentions.
+ * Participants should speak and act with good intentions, but understand that intent and impact are not equivalent.
  * Behaviour which can be reasonably considered harassment will not be tolerated.


ibiwan-intuit (approved):

hmdne:
I suggest we also remove this point from the conduct guidelines:

```
  * Participants will be tolerant of opposing views.
```

Some people may have views that when expressed, may be harmful to the interest of particular groups of people like big corporations. This has to be taken into account.

olivierlacan (approved):
I wholeheartedly approve this PR, for a host of reasons anyone willing to [educate](https://thoughtbot.com/blog/beyond-best-intentions) themselves could learn without me having to repeat them.

Intentions are not magical shields against criticism. Our community deserves our having higher standards for discourse and standing up to casual misogyny in the form of repeated tasteless so-called jokes by anyone, ever.

jacobherrington:
@hmdne I think that's a good point, but I'd like to be as explicit as possible in this PR; I watched a group of people use these specific phrases to justify making sexist remarks over a communication channel that falls under these guidelines. I really would like to see them removed, so I don't want to push beyond that goal.

However, I think it'd be a good idea for you to open another PR and suggest your change! I'd be in support of finding a better way to communicate the spirit of tolerance and mutual respect without implying that anyone is obligated to tolerate harmful behavior.

hmdne:
@jacobherrington I disagree. I view this PR is to fix the holes that people with other worldviews may use. Why fix one hole while leaving another wide open?

And while we are at it, maybe also we should correct this rule:

```
  * Behaviour which can be reasonably considered harassment will not be tolerated.
```

To become something like (what do other posters think? Can we make it even stricter?):

```
  * Behaviour which can be considered harassment will not be tolerated.
```

The law should always protect the victim, not the perpetrator. This is so that a harasser won't be able to use a meritocracy argument, which as we all know, is dehumanizing:

https://modelviewculture.com/pieces/the-dehumanizing-myth-of-the-meritocracy

jacobherrington:
@hmdne I don't disagree with your points, but I think these changes are distinct and for the sake of the Ruby maintainers, we should be specific in each PR.

I really do think having a broader discussion is important and your points should be part of that discussion.

Opening an issue on this repository could be a good place to start; re-writing the guidelines document will require buy-in from the larger community and the maintainers.

I'm hoping by keeping the suggestion small and pointing to a real world example of this language enabling harm, it won't be so hard to get the ball rolling.

hmdne:
I understand @jacobherrington. I will participate in this discussion. It will be very important to hear other feedback before we come up with a coherent PR.

Let me just quote the post, which started all this:

> Hello,
> 
> Another example - perhaps somebody could fix that (many errors have that
> source!)
> 
> Date.today
> 
> Date.today +1    # no error, but wrong result!
> 
> # Maybe this has been written for women, having calculated their age ;)

I am shocked that some people defended this person, as it is clearly against the universally accepted corporate values. There should be absolutely no tolerance for making jokes mentioning the protected classes.

JuanitoFatas (approved):
👍🏼 👍🏼 👍🏼

hmdne:
Encouraged by the quick success of this PR, I made #2691 with another small change. One step at a time at fixing the world :)

hsbt:
@JuanitoFatas WAIT. `en/conduct/index.md` is created by @matz originally. We should approve from him before merging this.

Try2Code:
> Encouraged by the quick success of this PR, I made #2691 with another small change. One step at a time at fixing the world :)

sry, but this is getting ridiculous. all this wording about good intentions  - so with this conduct we are all good ppl talking no trash, always the right way (yes, sir!). 

As a consequence, there will be 
- no more space for any kind of humor (cause someone could feel shocked)
- no more general criticism (cause some ppl might feel too offended) unless we use the perfect working for that (btw: do we have a dictionary already for expressing criticism in an correct way. I would like to have one - just to be on the save side)
-  but hang on: what if some ppl still feel offended even IF we use the right words? Do they loose the right to feel offended in that case?

zero tolerance means that we end up in our happy-pink-pony-farm-world and I don;t think this is a healthy development. I agree with the _try-to-be-nice_ approach to 100%, but it's only an approach. It's nothing we can enforce in my opinion. I think discussions should be open and free unless people get insulting like on a personal level. this is totally off-topic in should be done in a private chat. We need responsible ppl who moderate discussions so that everyone feels save, pointing to rules does not do the trick from my experience.

hmdne:
>but hang on: what if some ppl still feel offended even IF we use the right words? Do they loose the right to feel offended in that case?
@Try2Code if they are not a protected class, they will lose that right, yes. That's California Core Corporate Policies in work. And we enacted them.

Try2Code:
ok, so as long as u belong to a protected class you are allowed to feel offended. who are these classes?

hmdne:
@Try2Code That's common sense. We all know who are protected classes. For instance, members of board of directors of the big corporations are a protected class.

peterc:
Since California has been mentioned, I'm guessing these: https://www.senate.ca.gov/content/protected-classes

Everyone belongs to protected classes depending on context. It's not about some people being protected and some not. Everyone has a sexual orientation, marital status, race, etc. and harassment on such characteristics is verboten. (Although it was before too - the change in #2691 actually seems to _reduce_ the scope of what is considered harassment to me as it was universal before?)

hmdne:
@peterc I disagree. According to the latest guidelines, there is no such thing as a "reverse racism" or "reverse sexism".

Try2Code:
hang on (again): we use the term _protected class_ and do not know what is mean? semantics?

hmdne:
@Try2Code Everyone knows what a protected class means.

Try2Code:
sorry - this is a very weak explanation. esp. if you want to judge peoples behavior on it. THIS opens the door for many possibly fruitless discussions about _what people think_ should be a protected class.

Ether you have a definition for it or your don't. but in that case the term should simply not be used. 
Isn't it just to central for the whole point of having a conduct to do without a clear definition?

jacobherrington:
> @JuanitoFatas WAIT. `en/conduct/index.md` is created by @matz originally. We should get approval from him before merging this.

I would very much like @matz thoughts on this change. If he feels these guidelines shouldn't change it should be reverted.

However, I will definitely continue to advocate that these two guidelines change, even if it is decided that the commit is reverted.

There are likely more places the Ruby guidelines should be revisited, but these two lines were specifically used to justify making casual sexist jokes on the ruby-talk mailing list; they should be changed if Ruby, as a project, opposes that sort of behavior (I would think this so).

hmdne:
@Try2Code We don't have to define a "computer". We all know what protected classes are and if you don't, you will need to educate.

@jacobherrington Matz already approved of those changes.

Try2Code:
@hmdne you obviously do not want to answer my question. There IS a definition of a computer and a definition of a car. but here we do not touch the physical world, but social interaction between humans. 

And all you can say is: guy, you should know what I am talking about, but I won't tell you. please educate yourself. 

Then please, educate me! Where is the definition? I was just asking for that education in my post before, but you do not seem to willing to provide it. why? did I write something wrong?

hsbt:
Unfortunately, I also didn't know what means "protected classes" until this discussion. In Japan, We don't use "protected classes" terminology, We use "Race, Color, Religion, Gender, ... etc" explicitly.

I understand @Try2Code situation.  I'm asking this to @matz in next week.

PR #2692, opened 2021-10-01
Remove duplicate Ruby Essentials mention

TheStranjer:
Generally speaking, when you define something, you shouldn't mention the thing you're defining. Mentioning Ruby Essentials twice like this doesn't add anything to the writing, and in fact detracts from it.


JuanitoFatas (approved):
Great improvement, thanks!

PR #2698, opened 2021-10-08
Translate "Ruby 3.0.0 Preview 2 Released" (ko)

yous:
Translate #2559, #2561, #2563, #2564.

The actual diff: 355b880


yous:
Thanks!

PR #2714, opened 2021-10-29
Introduce TypeScript

stomar:
Set up TypeScript (add source directory and config file). The JavaScript files in `javascripts` can now be generated by running `tsc` in the root directory.

This might be useful for future changes of the JavaScript.
It does not affect the normal build process of the site.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -18,6 +18,7 @@ exclude:
   - lib
   - test
   - vendor
+  - tsconfig.json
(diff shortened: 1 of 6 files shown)


stomar:
This is kind of experimental and "just for fun"; but it doesn't add much overhead. Other config files (like e.g. package.json) etc. are not committed on purpose.

PR #2718, opened 2021-10-30
Use named link reference on user-groups page (translations)

stomar:
Also use https instead of http.


goz (commented):
ID looks good, thanks @stomar

PR #2719, opened 2021-10-31
Fix typos after spellchecker (ru)

razum2um:
I used https://github.com/hcodes/yaspeller to check all markdown files under `ru` folder

There are several locale-mismatched symbols (latin instead of cyrillic)

See how I discovered that: https://github.com/ruby/www.ruby-lang.org/pull/2720


lex111 (approved):

lex111:
@razum2um thanks!

PR #2722, opened 2021-10-31
Remove Google Analytics

stomar:
Remove the Google Analytics tracking code from the site.


stomar:
I do not see any benefits from using it, only concerns.

Waiting for reviews/opinions.

JuanitoFatas (approved):
I’m in favor of removal 👍🏼

Nakilon:
> I do not see any benefits from using it, only concerns.

I suppose it shows problems and traffic sources. If you don't want it to track you, you can install U-Block Origin.

stomar:
> If you don't want it to track you, you can install U-Block Origin.

I did not mean concerns as a visitor, but rather as a site owner; some points are made for example [here](https://dev.to/markosaric/why-you-should-remove-google-analytics-from-your-site-5c7h).

hsbt:
I'm considering to this now. Because GA provides statistics of new users of ruby lang to me.

hsbt:
OK, I decided to remove it from www.ruby-lang.org.

PR #2729, opened 2021-11-11
Fix tag name for Ruby 3.1.0-preview1

peterzhu2118:
Links are broken in the release page because the release tag is `v3_1_0_preview1` and not `ruby_3_1_0_preview1`.


ioquatix (approved):
LGTM.

PR #2768, opened 2021-12-25
Update tarball and documents

nurse:


stomar:
I assume `release.patch` was committed accidentally and removed it.

PR #2770, opened 2021-12-25
Fix: Incorrect description (en, ja)

sevenc-nanashi:
error_highlight is shown in NoMethodError, not NameError.

### DIFF
--- a/en/news/_posts/2021-12-25-ruby-3-1-0-released.md
+++ b/en/news/_posts/2021-12-25-ruby-3-1-0-released.md
@@ -61,7 +61,7 @@ test.rb:1:in `<main>': undefined method `time' for 1:Integer (NoMethodError)
 Did you mean?  times
 ```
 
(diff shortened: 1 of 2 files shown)


mame:
`NameError` is correct. Why do you think it is wrong?

```
$ cat test.rb
foo

$ ruby -v test.rb
ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]
test.rb:1:in `<main>': undefined local variable or method `foo' for main:Object (NameError)

foo
^^^
```

yous:
But the example in the article is `1.time {}`. Which one is the correct one?

mame:
@yous `NoMethodError` is a subclass of `NameError`.

yous:
Oh, now I understand.

sevenc-nanashi:
Sorry I didn't know it. I close it.

PR #2806, opened 2022-03-12
Fix cgi gem version in CVE-2021-41819 news post (translations)

meisyal:
I found another typo of `cgi` gem versions in CVE-2021-41819 news post. The versions are written like 0.2,1 and 0.1,1. The minor version and patch version are separated with comma. The correct versions should be 0.2.1 and 0.1.1.

This PR fixes the `cgi` versions.


gfvcastro (approved):
Thanks for your contribution!

PR #2820, opened 2022-04-07
Improve readability of Ruby 3.2.0-preview1 release notes

Aesthetikx:


diegoviola:
@Aesthetikx I think the part where it says:

    "The package name is different each platforms."

I would change that to "is different on each platforms" or similar.

https://github.com/ruby/www.ruby-lang.org/blob/bb09739d4f09d5b9aff81878183efc79ee29361e/en/news/_posts/2022-04-03-ruby-3-2-0-preview1-released.md?plain=1#L80

Thanks.

Aesthetikx:
Good input @diegoviola, I have updated that line as well.

PR #2840, opened 2022-04-21
Fix typos

Gao-Jun:


AhmadMajid:
pahse is also spelled phase.

Gao-Jun:
'pahse' -> 'phase' fixed and PR updated.

PR #2851, opened 2022-05-21
Fix FAQ: block params shadow outer lvars

nevans:
This is misleading: "that variable will be modified by the call to the block"

The outer variable isn't mutated; they are different variables.  The name points to one variable outside the block and a different variable inside the block. I'm not sure if my proposed change is better though! 😉

Also, I moved the "arguments are passed" paragraph to the top.  The original ordering was fine: it's reasonable to describe how to define block parameters first, and then how to pass arguments in them.  But it seems like reversing the order is the most direct answer to the question, as the question is worded.  Perhaps the original order was better, but we should add a second question heading: "How do I define a block's parameters"

### DIFF
--- a/en/documentation/faq/4/index.md
+++ b/en/documentation/faq/4/index.md
@@ -361,6 +361,9 @@ the method is invoked. It is evaluated using the scope of the method.
 
 ### How do I pass arguments to a block?
 
+Typically, arguments are passed to a block using `yield` (or an iterator
+that calls `yield`), or by using the `Proc.call` method.
+
 The formal parameters of a block appear between vertical bars at the start
 of the block:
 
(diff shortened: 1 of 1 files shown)


bo-tato:
I also got confused by this also when reading through the FAQ. I asked on reddit and [someone pointed out](https://old.reddit.com/r/ruby/comments/15i8f5m/proc_parameters/jv38j4e/) that before ruby 1.9.1 modifying a proc parameter would modify the outer variable of the same name, so it seems this section of the FAQ was written before then and hasn't been updated

en/documentation/faq/4/index.md
```diff
@@ -361,19 +361,20 @@ the method is invoked. It is evaluated using the scope of the method.
 
 ### How do I pass arguments to a block?
 
+Typically, arguments are passed to a block using `yield` (or an iterator
+that calls `yield`), or by using the `Proc.call` method.
```
nevans, line 365: I had previously just moved this paragraph to the top, but it should also be altered to use the appropriate syntax to indicate an instance method on Proc.  I prefer the `#` syntax, but this should use whichever is most consistent with other parts of this FAQ.

```suggestion
that calls `yield`), or by using the `proc.call` method.
```

```suggestion
that calls `yield`), or by using the `Proc#call` method.
```

PR #2852, opened 2022-05-24
Fix /en/documentation/faq/2/index.md

Abhideveloper28:
This statement looks wrong to me, It said a `A `def` statement outside of
a class definition become private methods of class Object`
that is not correct. I can call method (that define outside any class) to object of class Object.

### DIFF
--- a/en/documentation/faq/2/index.md
+++ b/en/documentation/faq/2/index.md
@@ -57,7 +57,7 @@ Ruby is a pure OO language that can masquerade as a procedural one. It has no
 functions, only method calls. In a Ruby method the receiver, also called
 `self`, is a hidden argument like `this` in C++. A `def` statement outside of
 a class definition, which is a function in Python, is actually a method call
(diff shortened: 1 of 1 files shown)


stomar:
The original statement seems to be correct:

~~~ ruby
irb(main):001:0> def my_test_method; end
=> :my_test_method
irb(main):002:0> Object.public_methods.grep /my/
=> []
irb(main):003:0> Object.private_methods.grep /my/
=> [:my_test_method]
~~~

stomar (changes requested):
Please clarify your statement _"I can call method (that define outside any class) to object of class Object"_ and what you did exactly.

jcsahnwaldt:
I think this can be closed without merging. The PR is almost two years old, OP didn't respond to questions, and the current version of the text seems to be correct.

hsbt:
The FAQ statement is correct. A top-level `def` defines a private method on `Object`, as shown above.

PR #2877, opened 2022-09-15
Translate Ruby 3.2.0-preview2 (es)

vtamara:


vtamara:
@gfvcastro @nurse  I translated to spanish on 16 september but it is still not visible in https://www.ruby-lang.org/es/

image

gfvcastro:
Good catch, @vtamara! Hey @hsbt, can you deploy the latest changes?

PR #2891, opened 2022-10-30
Update gem bundle

stomar:


stomar:
On my local machine the site does not build after these bundle updates...

stomar:
The site builds, but with extremely increased build time, both on CI and on my local machine - this needs investigating. With this build time, merging is no option.

Also, there are minor changes in the generated HTML, I'll look into those, too.

hsbt:
I did update the latest dependencies at https://github.com/ruby/www.ruby-lang.org/pull/3018

PR #2892, opened 2022-10-30
Translate "CVE-2021-41817" (ko)

yous:
Actual diff: 6cf099b


marocchino (approved):
👍🏽

stomar:
emojis in commit messages might be a problem in some terminals, I think it's better to stick to ASCII

PR #2904, opened 2022-11-21
Fix ruby-3-2-0-preview3-released links

yui-knk:


gfvcastro (approved):
Thanks for your contribution!

gfvcastro:
@yui-knk, can you update the commit message? Maybe it could be "Fix ruby-3-2-0-preview3-released links"; what do you think?

yui-knk:
@gfvcastro Thanks for your review. I updated the commit message. Could you take a look when you have time?

PR #2916, opened 2022-12-08
Added libyaml-dev for Psych 5.0

hsbt:


hsbt:
There is no dependency of Psych 5.0 in this repository. It's my mistake.

PR #2933, opened 2022-12-25
Translate Ruby 3.2.0 released (es)

vtamara:


vtamara:
@gfvcastro could you check please?

PR #2940, opened 2022-12-28
Improve release note of Ruby 3.2.0 (es)

gfvcastro:


vtamara:
Thank you @gfvcastro 
I agree with everything except translating "deprecated constants" as "constantes deprecadas".

According to https://dle.rae.es/deprecar the verb "deprecar" means "Rogar, pedir, suplicar con eficacia o instancia" that is not the meaning expected.

The translation suggested by https://translate.google.com/?q=deprecated%20constants is "constantes en desuso" que me parece mejor que lo que yo había escrito ("constantes despreciadas"), así que más bien use ese.

PR #2955, opened 2023-01-05
 Migrate Heroku to GitHub Pages

hsbt:
We use Heroku for www.ruby-lang.org because GitHub Pages didn't support HTTPS for custom domain in the past. Now, we can move to GitHub Pages with simple workflow. 

I try to migrate it.

Note: This work didn't contain domain configuration, only under the ruby.github.io domain.


naiff654321gmail:
gh pr checkout 2955

sorah (approved):

PR #2966, opened 2023-01-23
Update translation of documentation/installation (ja)

maimux2x:
## やったこと
https://www.ruby-lang.org/en/documentation/installation/#package-management-systems
上記英語版Rubyインストールページのパッケージ管理システムの箇所に合わせて、日本語版ドキュメントの説明が足りない箇所を修正。

PR #2971, opened 2023-01-25
Remove trailing whitespaces

gotoken:
This will resolve, for example, https://github.com/ruby/www.ruby-lang.org/actions/runs/3998473712/jobs/6861159757#step:6:46


sorah (approved):
follow-up: #2966

hsbt:
👍

Issue #2818, opened 2022-04-06
Follow up changes of upstream (ko) 2022-2024


yous:
@ruby/www-ruby-lang-org-i18n-ko We can keep track of the following list of pull requests or commits on upstream. See #1894.

- [ ] [en/privacy/index.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/privacy/index.md)
<details>
<summary>Translation Finished (TL;DR)</summary>

- [x] [2021-12-25-ruby-3-1-0-released.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2021-12-25-ruby-3-1-0-released.md) (#2781)
- [x] [2022-04-03-ruby-3-2-0-preview1-released.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2022-04-03-ruby-3-2-0-preview1-released.md) (#2817)
- [x] [2021-05-02-os-command-injection-in-rdoc.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2021-05-02-os-command-injection-in-rdoc.md) (#2821)
- [x] [2021-07-07-ruby-2-6-8-released.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2021-07-07-ruby-2-6-8-released.md) (#2856)
- [x] [2021-07-07-ruby-2-7-4-released.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2021-07-07-ruby-2-7-4-released.md) (#2856)
- [x] [2021-07-07-ruby-3-0-2-released.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2021-07-07-ruby-3-0-2-released.md) (#2856)
- [x] [2021-07-07-starttls-stripping-in-net-imap.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2021-07-07-starttls-stripping-in-net-imap.md) (#2856)
- [x] [2021-07-07-trusting-pasv-responses-in-net-ftp.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/news/_posts/2021-07-07-trusting-pasv-responses-in-net-ftp.md) (#2856)
[...]

PR #2982, opened 2023-02-11
Replace commit logs to GitHub releases (ko)

riseshia:
:link: #2818

Apply #2980 to ko.

Use `GitHub` as it is, I think it's better. 🤔
Should we translate to `깃헙`?


yous:
I think `GitHub` would be okay.

yous (approved):

riseshia:
🙏

PR #2985, opened 2023-02-27
Modify variants in ja

okuramasafumi:
I noticed that some variants in ja site is quite outdated and not informative. I also added truffleruby since it's still actively developed.


hsbt:
I agree to update them, But I don't agree to remove them from this page. Because we can still download and build them.

Can you mark outdated instead of removing from this page?

Note: rubinius is also outdated today.

okuramasafumi:
Thanks, I put them into a new H4 section.

hsbt (approved):

hsbt:
Thanks!

PR #3009, opened 2023-03-28
Use www.cve.org instead of nvd.nist.gov

hsbt:
We should use www.cve.org today.


hsbt:
> Hey @hsbt, do you think we should replace each URL that refers to nvd.nist.gov?

Yes. We should update each announce with this change.

PR #3034, opened 2023-04-11
Translate faq 7 (ko)

marocchino:


marocchino:
I'll merge it since there's no response. As always, I think it's good to have a separate PR for reviews.

PR #3042, opened 2023-04-18
Fix up typo "GiHub" to "GitHub"

riseshia:
small fix. Could you check this? @ruby/www-ruby-lang-org-editorial


JuanitoFatas (approved):
Thanks!

PR #3125, opened 2023-10-10
Install windows specific deps for Jekyll

nopeless:


nopeless:
https://github.com/jekyll/jekyll/issues/5935

nopeless:
should I update gem.lock as well

PR #3137, opened 2023-11-13
Translate Ruby 3.3.0-preview3 Released (zh_cn)

Gao-Jun:
As title


JuanitoFatas (approved):
Verified locally, LGTM.

CleanShot 2023-11-15 at 23 01 06@2x

Thanks for your contributions!

Issue #3134, opened 2023-10-25
www.ruby-lang.org needs a Privacy Policy page


lanrebr:
We find that our organization is refusing permission to install and use Ruby only because we cannot locate a privacy policy page that describes what what data is being collected by the web site, how it’s being collected, how it is stored, for how long it is stored , what it is being used for and how it is being shared.

Can we please add a Privacy Policy page nothing that the web site does not collect any information from any user?

hsbt:
@lanrebr Is https://www.php.net/privacy.php enough to your requirement? We are not company, foundation or NPO. It's difficult to provide detailed privacy policy.

lanrebr:
@hsbt thank you for looking into this. Yes I think something like the php privacy page would be a good starting point as it indicates that ruby will not need anything else than the email address and it would not share it with anyone.

hsbt:
Thanks. I'll consider that.

rubyFeedback:
Agreed. I think it may suffice to simply document what will happen with the data; and what data is stored. This should cover most of it; for the rest I think it would be fine to add a new issue at the ruby bug tracker, and to ask people to add their concern, along with a time limit for the discussion, say 1-2 months, so that it can then be closed and made official (and perhaps also have it for an upcoming ruby developer meeting to ask them about it).

lanrebr:
Thank you do very much!!!!

lanrebr:
Just a small note to let you know that there is a spell typo in new policy page. It reads Ascknowledgements instead of Acknowledgements.
Thank you

lanrebr:
Thank you

PR #3144, opened 2023-11-22
Added privacy policy for ruby-lang.org

hsbt:
Fixes https://github.com/ruby/www.ruby-lang.org/issues/3134

### DIFF
--- a/en/privacy/index.md
+++ b/en/privacy/index.md
@@ -1,0 +1,23 @@
+---
+layout: page
+title: "Privacy Policy for ruby-lang.org"
+lang: en
+---
+
+This privacy policy covers ruby-lang.org.
+
+## Email
+
+We will not give away your email address to anyone, who is not related to the operations of ruby-lang.org. We will also never ask you to send us any of your passwords via email.
+
+## Logfiles
+
+ruby-lang.org record access logs of the requests that reach the web servers, but we use those files for issue investigation and statistical purposes.
+
+## Cookies
+
+ruby-lang.org uses cookies to keep track of user preferences. Unless you login on the site, the cookies will not be used to store personal information and we do not give away the information from the cookies.
+
+## Ascknowledgements
+
+This privacy policy is based on the [php.net privacy policy](https://www.php.net/privacy.php).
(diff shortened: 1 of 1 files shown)


en/privacy/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+This privacy policy covers ruby-lang.org.
+
+## Email
+
+We will not give away your email address to anyone, who is not related to the operations of ruby-lang.org. We will also never ask you to send us any of your passwords via email.
+
+## Logfiles
+
+ruby-lang.org record access logs of the requests that reach the web servers, but we use those files for issue investigation and statistical purposes.
```
sorah, line 15:
```suggestion
ruby-lang.org records access logs of the requests that reach the web servers, but we use those files for debugging and statistical purposes.
```

sorah, line 15: As w.r-l.o uses GitHub Pages, it's worth to mention https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#data-collection

> www.ruby-lang.org uses GitHub Pages in its backend thus it is subject to GitHub's Data Collection. Refer to https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#data-collection for GitHub's privacy statement.

en/privacy/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Cookies
+
+ruby-lang.org uses cookies to keep track of user preferences. Unless you login on the site, the cookies will not be used to store personal information and we do not give away the information from the cookies.
```
sorah, line 19: > ruby-lang.org uses cookies to keep track of user preferences.

Do we? I don't think w.r-l.o has a user preference like https://www.php.net/my.php.

```suggestion
Some sites under ruby-lang.org (e.g. bugs.ruby-lang.org) uses cookies to keep track of user preferences. Unless you login on the site, the cookies will not be used to store personal information and we do not give away the information from the cookies.
```

sorah (changes requested):

en/privacy/index.md
hsbt: Yes, I noticed this for bugs.ruby-lang.org, ml.ruby-lang.org. I'll pick your suggestion.

nurse (approved):

PR #3146, opened 2023-12-04
[Data] Normalize all dates in releases.yml to Date instead of String

epistrephein:
In `_data/releases.yml`, 18 of the total 197 dates under the `date:` key are represented with quotes instead of without.

```ruby

yaml.find { |y| y['version'] == '3.2.0' }['date']
# => #<Date: 2022-12-25 ((2459939j,0s,0n),+0s,-Infj)>

yaml.find { |y| y['version'] == '3.2.2' }['date']
# => "2023-03-30"
```

This PR normalizes the 18 dates removing the quotes so that they're all parsed as `Date` class.

```ruby

# master
yaml.reject { |item| item['date'].is_a?(Date) }.count
# => 18

# this branch
yaml.reject { |item| item['date'].is_a?(Date) }.count
# => 0
```

### DIFF
--- a/_data/releases.yml
+++ b/_data/releases.yml
@@ -169,7 +169,7 @@
 # 3.2 series
 
 - version: 3.2.3
-  date: '2024-01-18'
+  date: 2024-01-18
   post: "/en/news/2024/01/18/ruby-3-2-3-released/"
   url:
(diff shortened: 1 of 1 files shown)


epistrephein:
Looks like v3.2.3 was also pushed with the release date as String instead of Date (ae9c2b0), I've pushed a fix for it as well.

Any chance to get this merged? @nurse @nagachika

PR #3148, opened 2023-12-04
Update hello_world.md for RuboCop

akirataguchi115:
Rubocop defaults complain about these double-quotes. Feedback and reasoning are appreciated.

### DIFF
--- a/en/examples/hello_world.md
+++ b/en/examples/hello_world.md
@@ -13,5 +13,5 @@ layout: null
 #
 # Here is the Code:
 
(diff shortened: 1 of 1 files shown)


hsbt:
We don't follow rubocop default rule.

akirataguchi115:
Thanks for the comment! Which formatting rules do we follow?

sorah:
We don't have a standard formatting rule, thus we can't acknowledge any formatting rule - accept changes based on those.

PR #3155, opened 2023-12-14
Fix typo on "Ruby 3.3.0-rc1 Released"

riseshia:
I think "compilation", not "copilation". 🤔 How do you think?


JuanitoFatas (approved):
Yes I think this is a typo based on https://github.com/ruby/ruby/blob/cfd91121ad71655ef9cea48450a0027a00a46f8f/NEWS.md?plain=1#L341, thank you for fixing this!

Great catch 💚

riseshia:
oh 😇

PR #3175, opened 2024-01-03
Add compatibility note about `Regexp::new`

MatheusRich:
I couldn't find any mentions of this either in the Ruby 3.3 release note or the 3.2 deprecation that @jeremyevans [mentioned](https://bugs.ruby-lang.org/issues/18797).

I'm not sure this should be considered a bugfix, so I added a small mention to this new behavior. LMK if anything else needs to change.


JuanitoFatas (approved):
Makes sense to me.

/cc @nurse

hsbt:
I have no opinion that.

sorah (approved):

PR #3185, opened 2024-01-18
 Translate "Ruby 3.2.3 Released" (ko)

riseshia:
:link: https://github.com/ruby/www.ruby-lang.org/issues/2818

Actual diff: 00561be72c673564ed7dc6a3332ba2a45a8d6e4d

PR #3186, opened 2024-01-18
Improve 2023-06-29-redos-in-uri-CVE-2023-36617 text (ko)

riseshia:
From: https://github.com/ruby/www.ruby-lang.org/pull/3185#discussion_r1457431845


yous (approved):
Thanks!

PR #3183, opened 2024-01-17
Adding Works on Arm logo on our English website.

junaruga:
This PR is to add Arm's logo sponsoring us for a Ubuntu Arm neoverse server (RubyCI - "Ubuntu arm neoverse") as their project Works on Arm to our English website: https://www.ruby-lang.org/en/about/website/ as our appreciation.

I got the logo image from "Arm Blue Logo" on [this page](https://www.arm.com/company/policies/trademarks/guidelines-corporate-logo). The url is https://www.arm.com/-/media/global/logos/Arm-logo-blue-pms313.svg . The image file is originally .svg file. I converted it to png file (image size 237 x 72 pixels).

You can check the updated document [here](https://github.com/junaruga/www.ruby-lang.org/blob/wip/website-sponsor-arm/en/about/website/index.md) on my forked repository.

I am going to confirm our folks at Arm before merging this PR.

PR #3191, opened 2024-02-03
Adding Works on Arm logo(ko)

riseshia:

Add arm logo to ko, too, which added on https://github.com/ruby/www.ruby-lang.org/pull/3183

Logo, which is actually "WORKS ON ARM", so It seems to be natural that we leaves the text as it is.
How do you think?

### DIFF
--- a/ko/about/website/index.md
+++ b/ko/about/website/index.md
@@ -36,7 +36,7 @@ lang: ko
 
 지원해주신 단체들에게도 감사드립니다.
 
-[Ruby Association][rubyassociation] (호스트)
+[Ruby Association][rubyassociation] (호스팅)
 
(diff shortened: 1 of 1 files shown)


ko/about/website/index.md
```diff
@@ -52,6 +52,10 @@ lang: ko
 
 <img src="../../../images/sponsor/heroku.png" alt="Heroku" width="216" height="256" />
 
+[Works on Arm][works-on-arm] (hosting)
```
yous, line 55: Can we also change '호스트' to '호스팅' in this page?

```suggestion
[Works on Arm][works-on-arm] (호스팅)
```

riseshia: sure~ fixed 9a0c60069

yous (approved):

riseshia:
🙏

PR #3207, opened 2024-03-23
Jekyll's default service port is 4000 instead of 9292 now

Gao-Jun:


JuanitoFatas (approved):
Thanks for the update!

hsbt:
👍

PR #3247, opened 2024-05-16
Update Japanese mirror to official URLs of Ring Server

yasuhirokimura:
Currently URLs of backend servers are listed. And some of them fail to be connected. So replace them with official URLs of Ring Server.


hsbt:
👍 Thanks

PR #3319, opened 2024-07-17
Update MailmanLists dashboard urls.

hsbt:
https://ml.ruby-lang.org/mailman3/postorius/lists/ruby-list.ml.ruby-lang.org/

is 404 status. It's updated with

https://ml.ruby-lang.org/mailman3/lists/ruby-core.ml.ruby-lang.org/


vtamara:
Change to file in spanish is correct.

PR #3338, opened 2024-08-05
Update ko/documentation/faq/3/index.md

riseshia:

Follow up https://github.com/ruby/www.ruby-lang.org/pull/3311


yous (approved):
Thanks!

PR #3355, opened 2024-08-27
Improve documentation page

st0012:
1. Remove dead or outdated links
    - Link 13 and 15 both point to https://ruby-doc.org/3.2.2, which is not
      up to date and don't provide more information than the official docs.
    - Link 17 points to https://rubydocs.org/ which is not maintained anymore.
    - Link 18 points to https://apidock.com/ only has information up to Ruby
      2.6, which is clearly outdated.
2. Make the link to VS Code directly visit its [Ruby doc](https://code.visualstudio.com/docs/languages/ruby)

### DIFF
--- a/en/documentation/index.md
+++ b/en/documentation/index.md
@@ -143,7 +143,7 @@ If you have questions about Ruby the
 [39]: http://kapeli.com/dash
 [docs-rlo-en]: https://docs.ruby-lang.org/en/
 [atom]: https://atom.io/
-[vscode]: https://code.visualstudio.com/
+[vscode]: https://code.visualstudio.com/docs/languages/ruby
 [eric]: https://eric-ide.python-projects.org/
 [kdevelop]: https://www.kdevelop.org/
 [kate]: https://kate-editor.org/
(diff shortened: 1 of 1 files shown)


JuanitoFatas (approved):
Looks good to me!

en/documentation/index.md
```diff
@@ -135,13 +118,8 @@ If you have questions about Ruby the
 [9]: http://www.ruby-doc.org/docs/ProgrammingRuby/
 [10]: http://pragmaticprogrammer.com/titles/ruby/index.html
 [12]: http://en.wikibooks.org/wiki/Ruby_programming_language
-[13]: http://www.ruby-doc.org/core
-[14]: https://ruby.github.io/rdoc/
-[15]: http://www.ruby-doc.org/stdlib
```
yous, line 140: Both `[13]` and `[15]` redirects to <https://ruby-doc.org/3.3.4/> now, so I think we don't have to remove these?

yous: `[34]` also links to <http://ruby-doc.org>, it needs to be https though.

PR #3375, opened 2024-09-24
[Data] Normalize all dates in releases.yml to Date instead of String

epistrephein:
In `_data/releases.yml`, 4 of the total 208 dates under the `date:` key are represented with quotes instead of without.

This PR normalizes the 4 dates removing the quotes so that they're all parsed as `Date` class.

```ruby
require 'yaml'

# master
yaml.filter_map { |i| i['version'] unless i['date'].is_a?(Date) }
# => ["3.2.5", "3.1.6", "3.1.5", "3.0.7"]
yaml.find { |i| i['version'] == '3.2.5' }['date']
# => "2024-07-26"


# this branch
yaml.filter_map { |i| i['version'] unless i['date'].is_a?(Date) }
# => []
yaml.find { |i| i['version'] == '3.2.5' }['date']
# => #<Date: 2024-07-26 ((2460518j,0s,0n),+0s,-Infj)>
```

(see #3146 for a similar fix some time ago)

### DIFF
--- a/_data/releases.yml
+++ b/_data/releases.yml
@@ -321,7 +321,7 @@
 # 3.2 series
 
 - version: 3.2.5
-  date: '2024-07-26'
+  date: 2024-07-26
   post: "/en/news/2024/07/26/ruby-3-2-5-released/"
   url:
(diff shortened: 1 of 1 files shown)


JuanitoFatas (approved):
Love the consistency, thanks!

PR #3382, opened 2024-10-07
Fix quoting of backtraces

eregon:
https://www.ruby-lang.org/en/news/2024/10/07/ruby-3-4-0-preview2-released/ renders it wrong, so I hope this fixes it.
Otherwise we might need to dedent the triple backticks completely.

Screenshot from 2024-10-07 20-09-35


eregon:
It works :)
Screenshot from 2024-10-07 20-13-56

PR #3388, opened 2024-10-09
Add neovim to Editors and IDEs section

nithinbekal:
Neovim is a popular choice for Rubyists, so I've added a link to it and to instructions to set it up with Ruby LSP.


hsbt:
Neovim has been listed on the documentation page since a17c89bff, and Ruby LSP is covered there for all the editors in the list, so I'm closing this.

PR #3390, opened 2024-10-09
Follow up "Update link for Programming Ruby book" (ko)

riseshia:

Follow up https://github.com/ruby/www.ruby-lang.org/pull/3385


yous:
Thanks!

PR #3417, opened 2024-11-05
Translate "Ruby 3.3.6 Released" (ko)

riseshia:
:link: #2818

Translates https://github.com/ruby/www.ruby-lang.org/pull/3416

Actual diff is a92c31a0e


riseshia:
🙏

PR #3250, opened 2024-05-20
Add rvm-windows to installation doc

magynhard:
Add the new RVM 4 Windows to the list of Ruby version managers at the installation documentation.

Included languages are `bg de en es fr id it ja ko pl pt ru tr zh_cn`

PR #3425, opened 2024-11-16
add rvm-windows to installation doc

magynhard:
Add the new RVM 4 Windows to the list of Ruby version managers at the installation documentation.

Included language is `en` only.

Note: As the [orignal Pull Request](https://github.com/ruby/www.ruby-lang.org/pull/3250) for all languages failed due inactive language reviewers, i decided to create one for English only.


JuanitoFatas (approved):
Thanks for all your work for RVM 4 Windows!!!

PR #3449, opened 2024-12-26
Use docs.ruby-lang.org for 3.4 release news

st0012:
With improved auto-linking, https://docs.ruby-lang.org/en/3.4/NEWS_md.html will be even more easier for readers to access updated classes/modules/methods from the page.

I used a fix url because existing `_data` attributes don't have the short-form version docs.ruby-lang.org uses.


st0012:
cc @hsbt

PR #3452, opened 2024-12-27
Fix broken link of 3.4.0 release news

Gao-Jun:
1. Add missing [Bug #20433] link
2. Fix sigstore.dev link. Previous link points to a relative link

BTW, the Japanese version may need a modification for sigstore link,
either remove the link at line 148, and modify the corresponding link at line 314
or remove the link line at 314.


hsbt:
👍

PR #3467, opened 2025-01-07
doc(de): fix links

bovi:
following #3452 to fix the links in the german news article


bovi:
@ruby/www-ruby-lang-org-i18n-de fixing the links as done in the other languages

PR #3512, opened 2025-02-18
Prepare preview pages

hsbt:
I'm preparing to use preview pages by cloudflare pages.

It uses only `BUNDLE_WITHOUT="development test"` for building pages. So, our current lockfile contained gems declared at development group. So, cloudflare pages couldn't resolve that dependencies in `bundle exec jekyll build`.


hsbt:
@ruby/www-ruby-lang-org You can check branch changes with cloudflare pages now. The preview url is provided at deployment page of GitHub like https://github.com/ruby/www.ruby-lang.org/runs/37377534758.

PR #3551, opened 2025-04-18
Translate "Ruby 3.5.0 preview1 Released" (ko)

riseshia:
:link: #3461 

Translation of https://github.com/ruby/www.ruby-lang.org/pull/3549

Actual diff is ecd115f1010c8f5fe5d250a4ae6b8e4c227bb645

PR #3553, opened 2025-04-18
Apply translation review to "Ruby 3.5.0 preview1"

riseshia:
Follow up https://github.com/ruby/www.ruby-lang.org/pull/3551


yous (approved):
Thanks!

PR #3568, opened 2025-05-14
clean up

whitehorsesoft:
Remove outdated and broken links to other Ruby implementations. I went to each of the links, they either link to projects no longer actively supported or projects that haven't been maintained for years.


zenspider:
This looks good to me. Bump

hsbt:
It would be better to separate the links into sections and keep them as archives.

Issue #3461, opened 2025-01-01
Follow up changes of upstream (ko)


riseshia:
@ruby/www-ruby-lang-org-i18n-ko We can keep track of the following list of pull requests or commits on upstream. See #2818.

- [ ] [en/privacy/index.md](https://github.com/ruby/www.ruby-lang.org/blob/master/en/privacy/index.md)

<details>
<summary>Translation Finished (TL;DR)</summary>

- [x] #3445 (#3446)

</details>

riseshia:
#3456 #3432 #3427

yous:
- #3452

riseshia:
https://github.com/ruby/www.ruby-lang.org/pull/3464/files/08de1246d5cac1970eccb06ef026a4d478f7d811#diff-06b5ba766f2ef1f044b32f84e71e0a1687a7ea102961b74908186493e30894a3

riseshia:
#3471

riseshia:
https://github.com/ruby/www.ruby-lang.org/pull/3484

riseshia:
#3455

riseshia:
#3489

riseshia:
#3502

riseshia:
#3516

riseshia:
#3513

yous:
Maybe #3212?

riseshia:
Maybe....

riseshia:
II added #3513 without much thought, but do we really need to translate news that only listed in the ja? :melting_face:

marocchino:
I asked this before and was told that local events do not need to be translated.

riseshia:
✅

yous:
- #3529
- #3530

yous:
- #3536

yous:
- #3542

riseshia:
#3549

riseshia:
https://github.com/ruby/www.ruby-lang.org/pull/3560

yous:
- #3567

riseshia:
https://github.com/ruby/www.ruby-lang.org/commit/e7c946a04e342b4ffac1b58479dde7655fa12d8c

yous:
- #3590

yous:
- #3595

riseshia:
- #3596

yous:
- #3614

riseshia:
#3617

riseshia:
[...]

PR #3577, opened 2025-05-28
Follow up "Use docs.ruby-lang.org instead of ruby-doc.org"

riseshia:

Follow up #3576


yous (approved):
Thanks!

PR #3579, opened 2025-06-02
Follow up "Use docs.ruby-lang.org instead of ruby-doc.org"

timfanda35:
Follow up https://github.com/ruby/www.ruby-lang.org/pull/3576


bugtender (approved):
Lovely! thanks @timfanda35

PR #3615, opened 2025-09-17
Sync BSD-2-Clause license

syleeeee:
I found that the license is different from ruby/ruby.
https://github.com/ruby/ruby/blob/master/BSDL

Which one is supposed to be?


kou:
Ruby is dual license of 2-clause BSDL and its original license you removed.
We must not remove the original license part.

See also the part you didn't remove:

> You can redistribute it and/or modify it under either the terms of the
> 2-clause BSDL (see the file BSDL), or the conditions below:

syleeeee:
Oh, I didn't see `or`. Let me close this PR. Thank you.

PR #3616, opened 2025-09-17
Translate "Ruby 3.4.6 Released" (ko)

riseshia:
:link: #3461

Translates https://github.com/ruby/www.ruby-lang.org/pull/3614

Actual diff is 889d931

PR #3618, opened 2025-09-17
Fix "Ruby 3.4.6 Released" (ko)

riseshia:
Apply review from https://github.com/ruby/www.ruby-lang.org/pull/3616#pullrequestreview-3232802205


yous (approved):
🙇

PR #3639, opened 2025-10-10
Add Open Ruby Discord Server

jclusso:
Add the Open Ruby Discord Server that welcomes differing opinions.

PR #3654, opened 2025-10-22
remove discord from community page

gregmolnar:
The Ruby Discord community isn’t limited to the one linked from the community page and there are plenty of other ones. The listed one isn't official, so in the name of fairness, I don't think it should be promoted on this website, unless we add all of them and people can freely choose which one they would prefer to join.


L-A:
I'll say it at the top to save everyone's time – this PR is fishing for victimhood.

schneems:
This is a continuation of the debate that was locked on #3639. Locking and closing this as well.

PR #3666, opened 2025-10-31
Restore missing link to Ruby Central in the "Ruby Conferences"

ablzh:
image


schneems:
Looks like it was removed by mistake in this PR https://github.com/ruby/www.ruby-lang.org/pull/3275/files. Thanks!

If there's a markdown linter that would detect links missing a URL, that could be a nice addition (pending complexity/overhead).

PR #3672, opened 2025-11-10
[Data] Normalize all dates to Date instead of String

epistrephein:
In `_data/releases.yml`, 3 of the total 229 dates under the `date:` key are written with quotes instead of without quotes.

This PR normalizes the 3 dates removing the quotes so that they're all parsed as `Date` class.

```ruby
require 'yaml'

# master
yaml.filter_map { |i| i['version'] unless i['date'].is_a?(Date) }
# => ["3.3.10", "3.3.9", "3.2.9"]
yaml.find { |i| i['version'] == '3.3.9' }['date']
# => "2025-07-24"


# this branch
yaml.filter_map { |i| i['version'] unless i['date'].is_a?(Date) }
# => []
yaml.find { |i| i['version'] == '3.3.9' }['date']
# => #<Date: 2025-07-24 ((2460881j,0s,0n),+0s,-Infj)>
```

(see #3566, #3375 and #3146 for the same fix some time ago)

### DIFF
--- a/_data/releases.yml
+++ b/_data/releases.yml
@@ -339,7 +339,7 @@
 # 3.3 series
 
 - version: 3.3.10
-  date: '2025-10-23'
+  date: 2025-10-23
   post: "/en/news/2025/10/23/ruby-3-3-10-released/"
   url:
     gz: https://cache.ruby-lang.org/pub/ruby/3.3/ruby-3.3.10.tar.gz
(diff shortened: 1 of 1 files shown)


hsbt:
We need to resolve this root causes. The release manager used https://github.com/ruby/actions/actions/runs/18744734895/job/53469239723#step:5:278 for copy-and-paste. That provides single quote for release date.

kpumuk:
Would it make sense to add a check to the linter? https://github.com/ruby/www.ruby-lang.org/pull/3668

PR #3676, opened 2025-11-18
Add missing footnote to "Ruby 4.0.0 preview2 Released"

Gao-Jun:
As title.

PR #3681, opened 2025-11-20
Translate Ruby 4.0.0 preview2 Released (zh_tw)

timfanda35:
Translate Ruby 4.0.0 preview2 Released (zh_tw)

Thank you.


schneems:
The version was missing links, they've just been added via https://github.com/ruby/www.ruby-lang.org/pull/3676 can you add them to your translation please?

PR #3700, opened 2025-12-20
Use CRuby 3.4 to run example codes

y-yagi:
TryRuby supports CRuby 3.4. So let's use the latest version. Ref: https://github.com/ruby/TryRuby/pull/216


hsbt:
We removed that from https://github.com/ruby/www.ruby-lang.org/pull/3699

PR #3701, opened 2025-12-20
Remove engine pin for TryRuby

okuramasafumi:
https://github.com/ruby/TryRuby/blob/master/app/try_ruby.rb#L19

TryRuby now uses Ruby `3.4.1` as a default engine. I didn't see a good reason to use older Ruby, `3.3.0` here, so I removed that part.


koic:
It appears to be a duplicate of the already open PR #3699.

PR #3703, opened 2025-12-20
Don't mutate a string literal in main code sample

byroot:
I know this is still being debated, but as of the current version of Ruby, the first code sample does emit a deprecation warning (hidden by default but still).

I think it's would be better to not mutate a string literal.

### DIFF
--- a/bg/examples/i_love_ruby.md
+++ b/bg/examples/i_love_ruby.md
@@ -8,7 +8,7 @@ say = "I love Ruby"
 puts say
 
 # Output "I *LOVE* RUBY"
-say['love'] = "*love*"
+say = say.sub("love", "*love*")
 puts say.upcase
 
 # Output "I *love* Ruby"


BurdetteLamar:
@byroot, I don't feel qualified to review.

byroot:
@BurdetteLamar I ddin't make the request, it's automatic.

But @hsbt is in the reviewer, I think he can make that call.

BurdetteLamar:
Okay, thanks.

bg/examples/i_love_ruby.md
```diff
@@ -8,7 +8,7 @@ say = "I love Ruby"
 puts say
 
 # Output "I *LOVE* RUBY"
-say['love'] = "*love*"
+say = say.sub('love', "*love*")
```
st0012: nit: I think this is the only place in the examples using single quotes. Let's use double quotes too for consistency and avoid confusion?

```suggestion
say = say.sub("love", "*love*")
```

PR #3713, opened 2025-12-22
Added CSS build check by Actions

hsbt:
This is remaining work of redesign project. We should check to css build every PRs.


hsbt:
I'm not sure why `npm run build-css` at GitHub Actions generate the following diff:

```css
+.rounded{
+  border-radius: 0.25rem;
+}
+
 .rounded-2xl{
   border-radius: 1rem;
 }
@@ -4567,6 +4571,10 @@ body:is(.dark *){
   text-decoration-line: underline;
 }
...
```

PR #3717, opened 2025-12-22
Inline code examples at build time instead of fetching via JavaScript

### LINKED ISSUE #3715: Syntax highlighting not displaying in some cases
I've observed cases where code syntax highlighting occasionally fails to display properly.

## Possible causes:

* Browser extensions may be interfering with or modifying the JavaScript (e.g., there may be a conflict with the 1Password extension)
* Some users report that syntax highlighting appears after refreshing the browser twice
* Removing the JS fetch for code examples on the top page (which was unnecessary in the first place 🙏) might resolve this issue

## Environment details:

If you're experiencing this issue, please share:

- Browser and version
- Active browser extensions
- Steps to reproduce

Image

JuanitoFatas:
This PR improves loading examples to not depend on needing JavaScript. Mostly done by Claude. Fixes #3715.

## Problem

Code examples were loaded via `fetch()` requests at runtime, which could fail due to:

- Browser extensions (1Password, ad blockers) interfering with JavaScript
- Network issues or timing problems
- JavaScript execution failures

## Fix

Inline the syntax-highlighted code directly in the HTML during Jekyll build, eliminating the need for runtime fetch requests.

## Changes

- Created `_includes/home/code_examples/*.html` with the three Ruby examples
- Modified `_includes/home/try_ruby.html` to use Jekyll includes instead of empty placeholders
- Simplified `javascripts/try-ruby-examples.js` - removed fetch logic, now only generates TryRuby button URLs from existing DOM content
- Simplified `javascripts/hero-animation.js` - removed async loading dependency
- Added `<noscript>` fallback in `_layouts/homepage.html` for graceful degradation when JavaScript is disabled/unavailable

## Result

- No more fetch requests to `/examples/` endpoints
- Syntax highlighting displays immediately (no JS dependency for content)
- Works even with JavaScript disabled
- Faster page load (3 fewer HTTP requests)

## Verifications

- [x] I have verified example loads with syntax highlights 👍🏻 

#### No need to `fetch`

CleanShot 2025-12-22 at 15 33 51@2x

CleanShot 2025-12-22 at 15 54 36@2x

#### When JavaScript is disabled

CleanShot 2025-12-22 at 15 34 33@2x

### DIFF
--- a/stylesheets/compiled.css
+++ b/stylesheets/compiled.css
@@ -1962,6 +1962,16 @@ body:is(.dark *){
   color: rgb(250 250 249 / var(--tw-text-opacity, 1));
 }
 
+/* Ensure syntax highlighting colors are not overridden by body color */
+
+.highlight code {
+  color: inherit;
+}
+
+.highlight pre {
+  color: var(--color-text-default);
+}
+
 .container{
   width: 100%;
   margin-right: auto;
--- a/stylesheets/tailwind.css
+++ b/stylesheets/tailwind.css
@@ -23,4 +23,12 @@
     @apply bg-white text-stone-900 text-base leading-5 font-default;
     @apply dark:bg-stone-900 dark:text-stone-50;
   }
+
+  /* Ensure syntax highlighting colors are not overridden by body color */
+  .highlight code {
+    color: inherit;
+  }
+  .highlight pre {
+    color: var(--color-text-default);
+  }
 }


hsbt:
Should we also remove `en/examples/*.md`? (It's available on all languages)

JuanitoFatas:
> Should we also remove `en/examples/*.md`? (It's available on all languages)

How about we ship this CSS fix PR first, verified we dont need these examples, then another PR to remove all examples? What do you think? :)

hsbt:
Ah, make sensed. We may use that examples from other place.

JuanitoFatas:
cc @taea would love your review on this 🙇🏻‍♂️

taea:
@JuanitoFatas Thank you so much!! I'll check on that!

taea:
@JuanitoFatas LGTM! :+1:  Tested locally and it works perfectly ✨  Thank you so much!

JuanitoFatas:
Thanks for the review @taea @hsbt!!

Tested this works on production with and without JavaScript 🎉 

https://github.com/user-attachments/assets/97edfa02-6210-4737-99d0-e2e13b00c184

PR #3718, opened 2025-12-22
Improve font family with add lang attr

liberaldev:
I added appropriate `lang` attributes based on language, adjusted the `font-family` accordingly, and removed system-ui from the `font-family` to eliminate a factor that reduces readability in CJK environments.

note: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-family#system-ui


_layouts/root.html
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
+{% if page.lang != null %}
+<html lang="{{ page.lang | to_html_lang }}">
+{% else %}
+<html>
+{% endif %}
+  {{ content }}
+</html>
```
sorah: https://thoughtbot.com/blog/no-newline-at-end-of-file

sorah:
thanks!

PR #3723, opened 2025-12-22
Fix PowerShell language name for Rouge

eagletmt:
`highlight ps1` is intended to syntax highlight as PowerShell I think, but the correct language name in Rouge is powershell.
https://github.com/rouge-ruby/rouge/wiki/List-of-supported-languages-and-lexers


riseshia:
❤️ thanks!

PR #3724, opened 2025-12-22
Fix winget installation instructions (ja)

eagletmt:
This PR applies the improvements from #3126 to the ja version.


hsbt:
👍 Thanks!

PR #3730, opened 2025-12-23
Add missing footnote to "Ruby 4.0.0 preview3 Released"

Gao-Jun:
As title.
All three translated languages(en, ja, ua) are fixed.
It seems that the es translation doesn't translate the corresponding content, so skip it.


hsbt (approved):
👍

Issue #3731, opened 2025-12-23
The code for recently added Ukrainian language is wrong, and should be "uk", not "ua"


vdoto4554872:
https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes

Moreover, already existing Korean (ko) and Japanese (ja) correctly use language codes, and not the country ones of "kr" and "jp" respectively. "ua" is the country code, not the language one.

afurm:
Thanks. I will take care.

PR #3733, opened 2025-12-23
Rename Ukrainian localization from ua to uk

afurm:
Renames the Ukrainian localization to use the correct ISO language code uk instead of the country code ua. This aligns it with other localizations like ko (Korean) and ja (Japanese) that correctly use language codes.

Fixes: https://github.com/ruby/www.ruby-lang.org/issues/3731


index.html
```diff
@@ -13,8 +13,8 @@
         "pt":    "pt",
         "ru":    "ru",
         "tr":    "tr",
-        "uk":    "ua",
-        "ua":    "ua",
+        "uk":    "uk",
+        "ua":    "uk",
```
hsbt, line 17: Should we remove "ua" from this list?

afurm: Yes, thanks i will remove.

afurm: Done. https://github.com/ruby/www.ruby-lang.org/pull/3734

PR #3729, opened 2025-12-23
Translate 3.4.8 and 4.0.0_preview3 release(zh_cn)

Gao-Jun:
As title.

PR #3735, opened 2025-12-23
Fix stdlib gem versions for 4.0.0-preview3

nevans:
Some of the gem versions in the 4.0.0-preview3 announcement were incorrect.  I double checked the versions, and the [`NEWS.md`](https://github.com/ruby/ruby/blob/v4.0.0-preview3/NEWS.md?plain=1#L266-L344) file in the ruby repository's `v4.0.0-preview3` tag does appear to be correct.  So I copied that entire section of the `NEWS.md` file into the release announcements for the en, ja, uk, and zh_cn languages (the es announcement does not seem to include stdlib updates).


nevans:
Rebased, and updated to include #3729.

hsbt (approved):

PR #3757, opened 2025-12-25
Implement English fallback for missing translations to prevent 404 errors

liberaldev:
Previously, users needed to manually switch languages to view these pages. This became especially inconvenient with newly published content—users had to wait for translation PRs to be merged or manually change their language settings to access the information. Furthermore, attempting to switch a post's language to one without an existing translation triggered a 404 error.

This PR addresses these issues by introducing a fallback mechanism. Now, if a specific translation is unavailable, the English version of the page is displayed instead of returning an error. As a result, newly published content is immediately accessible to all users, regardless of their language settings.

## Changes

* Added logic to check for translation availability.
* Configured the system to render the English page as a fallback when the target language content is missing.

---
**Attribution**
The code in this pull request was primarily generated using JetBrains Junie


hsbt:
Instead of keeping full copies of English pages under every language, #4008 now redirects to the English page whenever a translation is missing, so I'm closing this in favor of that.

PR #3771, opened 2025-12-28
Improve font stack

liberaldev:
This pull request refines how CJK (Chinese, Japanese, Korean) font stacks are applied throughout the site, improving language-specific font rendering and preventing CJK font styles from being incorrectly applied to non-CJK content. It introduces a new CSS variable for the CJK font stack and restructures selectors to ensure fonts are only applied where appropriate.

## Font Stack Improvements

* Introduced a new CSS variable `--font-sans-cjk` for the CJK font stack in `variables.css`, centralizing the font definition for easier maintenance. [[1]](diffhunk://#diff-5a37150576341ee5a1af874c9e7e935051ecc37c1f6d1992c6605a86093dc499R72-R74)

## Selector and Specificity Fixes (Improvement of #3744)

* Updated selectors to ensure CJK fonts are only applied to elements and languages that require them, preventing CJK font styles from leaking into non-CJK content. The following elements and classes are now explicitly excluded from CJK font application:
  * Monospace semantic elements: `code`, `pre`, `kbd`, `samp`
  * Font utility classes: `.font-default`, `.font-sans` (when used without CJK language attributes)
  
  This ensures that code blocks and explicitly styled elements maintain their intended fonts even when the page language is set to CJK. [[1]](diffhunk://#diff-b1f439837aa6bff143be7147ea518f4b185e3f432122043ada0d8da5c803a455R22-R46)

* Moved the CJK font application rules for `.font-default` and `.font-sans` classes to a new utility layer in `tailwind.css` to ensure proper specificity. These rules now apply the CJK font stack specifically when used with CJK language attributes (`lang="ja"`, `lang="ko"`, `lang="zh-CN"`, `lang="zh-TW"`), providing proper font rendering for elements that explicitly declare CJK languages. [[1]](diffhunk://#diff-b1f439837aa6bff143be7147ea518f4b185e3f432122043ada0d8da5c803a455R57-R69)

## Summary

These changes collectively ensure that CJK font stacks are applied accurately and only where intended, improving typographic consistency for multilingual content. The improved selector specificity resolves issues where CJK fonts would override non-CJK content, while the new utility layer ensures proper font rendering for elements with explicit CJK language declarations.


hsbt:
I built the site with and without this change and compared the computed fonts. Master already renders each name in the language selector in its own font on every page, while with this change they follow the page language instead, so 한국어 gets Noto Sans JP on /ja/ and no CJK font on /en/. Since the case described is already handled, I'm closing this.

PR #3893, opened 2026-04-06
downloads pages: French translation

noraj:


etiennebarrie (approved):
I pushed a few small changes, let me know if that's ok.

noraj:
Thanks, looks ok to me.

PR #3943, opened 2026-05-20
Translate "Ruby 4.0.5 Released" & "CVE-2026-46727" (ko)

riseshia:
:link: https://github.com/ruby/www.ruby-lang.org/issues/3461

Translation of: #3942

Actual diff is: e80deb354, 2eaee1ec8

PR #3945, opened 2026-05-20
Fix wording in CVE-2026-46727 announcement (ko)

riseshia:
From https://github.com/ruby/www.ruby-lang.org/pull/3943#discussion_r3271974009


yous (approved):
Thanks!

PR #3950, opened 2026-05-30
Extract hardcoded homepage strings to locale files (Fixes #3802)

ablzh:
This PR resolves #3802 by extracting the remaining hardcoded English strings on the homepage into the locale system, enabling full localization for non-English users.

Specifically, the following strings were extracted to _data/locales/home/en.yml:
   1. The "A Programmer's Best Friend" tagline in the Hero section and SEO meta tags (hero.tagline).
   2. The "TRY!" button text (try_ruby.button_text).
   3. The "Try Ruby" link text at the bottom of the section (try_ruby.bottom_link_text).

**Implementation Details & Fallback Mechanism**
To avoid breaking the homepage for languages that do not yet have these translations, I implemented a strict string-level English fallback using the Liquid default filter:

```
   1 {% assign en_home = site.data.locales.home.en %}
   2 {{ home.hero.tagline | default: en_home.hero.tagline }}
```

**Instructions for Locale Maintainers**
Other language maintainers do not need to do anything immediately; their pages will continue to display the English strings safely.

To localize these strings for your language, simply add the following keys to your respective _data/locales/home/*.yml file:

```
   1 hero:
   2   # ...
   3   tagline: "Your translation here"
   4
   5 try_ruby:
   6   # ...
   7   button_text: "Your translation here"
   8   bottom_link_text: "Your translation here"
```

PR #3952, opened 2026-06-02
Follow to use locale config on ko

riseshia:

Follow up https://github.com/ruby/www.ruby-lang.org/pull/3950


yous (approved):
Thanks!

PR #3964, opened 2026-06-30
Translate "Ruby 3.4.10 Released" (ko)

riseshia:
Translation of: #3963
Actual diff is: d43f06497


yous (approved):
Thanks!

PR #3965, opened 2026-07-03
Fix broken NEWS link in Ruby 4.0.0 release post

ErnaneJ:
The NEWS link in the [4.0.0 release post](https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/) points to `https://docs.ruby-lang.org/en/v4.0.0/NEWS_md.html`, which returns 404.

The template `{{ release.tag }}` resolves to `v4.0.0`, but docs.ruby-lang.org uses a `major.minor` path without the `v` prefix. The correct URL is `https://docs.ruby-lang.org/en/4.0/NEWS_md.html`, as confirmed by @hsbt in the issue.

Fixes #3764

PR #3971, opened 2026-07-06
Fix NEWS link for Ruby 4.0.0 release notes (ko)

riseshia:

Follow up #3965


marocchino (approved):

yous (commented):
Thanks!

PR #4011, opened 2026-09-18
Translate "Ruby 4.0.7 Released" (ko)

riseshia:

Translation of #4007


yous (approved):
Thanks!

